PR #3, opened 2020-11-01
fix typos in FastAPIWrapper

sinisaos:
These is just minor documentation fixes to prevent SyntaxError for usage of FastAPIKwargs in FastAPIWrapper.


dantownsend:
Perfect - thanks for this. I must've had one too many beers when writing those docs!

sinisaos:
It’s never too much beers :). Cheers

PR #4, opened 2020-12-06
Adding readable columns to single row

sinisaos:
Maybe it be useful to have readable column for FK in representation of single row like we have in representation of all rows in PiccoloCRUD. I'm sorry if that functionality already exist but I can't find it.

### DIFF
--- a/piccolo_api/crud/endpoints.py
+++ b/piccolo_api/crud/endpoints.py
@@ -152,6 +152,7 @@ class PiccoloCRUD(Router):
         return create_pydantic_model(
             self.table,
             include_default_columns=True,
+            include_readable=True,
             model_name=f"{self.table.__name__}Output",
         )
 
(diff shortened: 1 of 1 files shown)


dantownsend:
@sinisaos Yes, this is a great suggestion. I had to refactor ``PiccoloCrud`` slightly, so that the ``_get_single`` method has access to the HTTP query parameters. This way, the behaviour can be enabled by doing `/mytable/1/?__readable=true`. What do you think?

I reused your code, but put it in a different branch, as the changes are easier to show in code than describe in words:

https://github.com/piccolo-orm/piccolo_api/pull/5

Thanks :)

sinisaos:
@dantownsend Great. That's exactly what I mean and your version is much better :) . You can close this pull request.

dantownsend:
Closing - used as basis for https://github.com/piccolo-orm/piccolo_api/pull/5

PR #8, opened 2020-12-14
adding auth example to FastAPIWrapper

sinisaos:
@dantownsend After our conversation I'm adding auth example to FastAPIWrapper. Hope you like it.

### DIFF
--- a/docs/source/fastapi/index.rst
+++ b/docs/source/fastapi/index.rst
@@ -96,8 +96,7 @@ We also mark one of the endpoints as deprecated.
 Authentication
 --------------
 
-If you want protect some endpoints using ``FastAPIKwargs``,
-you can pass auth dependencies into ``FastAPIApp.add_api_route``.
+You can use FastAPI's `OAuth2 <https://fastapi.tiangolo.com/tutorial/security/>`_  features to protect some endpoints.
 
 Example
 ~~~~~~~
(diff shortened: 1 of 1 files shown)


docs/source/fastapi/index.rst
```diff
@@ -93,6 +93,46 @@ We also mark one of the endpoints as deprecated.
         )
     )
 
+Authentication
+--------------
+
+If you want protect some endpoints using ``FastAPIKwargs``,
+you can pass auth dependencies into ``FastAPIApp.add_api_route``.
```
dantownsend, line 100: I'd maybe change this sentence to the following:

```
You can use FastAPI's `OAuth2 <https://fastapi.tiangolo.com/tutorial/security/>`_  features to protect some endpoints.
```

The only reason I mention this is because there are other ways to protect the endpoint - for example, using ASGI middleware. This is FastAPI's way. What do you think?

docs/source/fastapi/index.rst
```diff
@@ -93,6 +93,46 @@ We also mark one of the endpoints as deprecated.
...
+        ),
+        fastapi_kwargs=FastAPIKwargs(
+            all_routes={'tags': ['Task']},  # Added to all endpoints
+            post={"dependencies": [Depends(oauth2_scheme)]}, # protected route
+            put={"dependencies": [Depends(oauth2_scheme)]},  # protected route
+            patch={"dependencies": [Depends(oauth2_scheme)]},  # protected route
+            delete_single={"dependencies": [Depends(oauth2_scheme)]},  # protected route
+        )
+    )
+
```
dantownsend, line 135: Do you think it would be useful to link to your example project here? For example:

```
See this `example project <https://github.com/sinisaos/headless-forum-fastapi>`_ for more details.
```

Because your project also includes the login endpoint etc.

sinisaos:
I agree with you and I make proposed changes.

dantownsend:
Looking good 👍

PR #26, opened 2021-02-11
Missing new endpoint

sinisaos:
After adding FastAPI as backend `/table/new` endpoint return `404` error. Adding missing `new` endpoint for creating rows in Piccolo Admin.


dantownsend:
@sinisaos Good catch! Moving it to FastAPI was definitely more hassle than I imagined.

dantownsend:
@sinisaos I've released the fix. Thanks for that.

I'm just writing a few tests for cursor pagination, but should release that soon.

sinisaos:
@dantownsend There were a few problems, but I think everything is fine now. In general, I like Starlette more because it’s easier. How do you like cursor pagination for now? Is that how you imagined it?

dantownsend:
@sinisaos I like the cursor pagination stuff. I'm thinking of changing slightly so the cursor returned is the last element in the list, rather than the next value, if that makes sense.

I should finish writing the tests this weekend and will get it merged in.

sinisaos:
@dantownsend Great. If the result is the same then it doesn't matter.

I tried to emulate way from [article](https://slack.engineering/evolving-api-pagination-at-slack/) wich you posted, and they use last result which isn’t returned in the result set, but use that value as the `next_cursor`, which is first result in next result set. Then you must use `id >= next_cursor`

dantownsend:
@sinisaos Yeah, I guess that is the most correct way of doing it.

PR #27, opened 2021-03-01
the get_ids endpoint can now be filtered using the search parameter

dantownsend:
The `get_ids` endpoint of `PiccoloCRUD`, accessed via `/tablename/ids/`, now accepts a search parameter.


dantownsend:
This functionality is to support a search widget in the Piccolo Admin: https://github.com/piccolo-orm/piccolo_admin/pull/36

PR #35, opened 2021-03-24
Cursor pagination

dantownsend:


sinisaos:
@dantownsend I tried your concept and it is very interesting, but the correct result on the ASC forward I can only get from next_cursor. On ASC forward, the previous_cursor does not return previous page rows but rows of the first page. DESC does not work in either direction. My attempt is based on [this](https://medium.com/@rcls/how-to-implement-slacks-cursor-based-pagination-spec-with-dynamodb-3ed08ef12297) and for me this concept is the easiest to follow. Other cursor implementations are too complicated for me. After a lot of trying I was finally able to make a solution that works both ways.

dantownsend:
@sinisaos Yeah, my implementation is broken at the moment - I didn't get around to finishing it. I just uploaded it so you can take a look.

It's a very tricky problem. I'll have a read of that article you linked to.

sinisaos:
@dantownsend Great. It is basically a Slack implementation (only one cursor) with the ability to change direction through `dir` (I use `__previous`) parameters. To me, this implementation was the easiest to understand.

Issue #48, opened 2021-06-27
Use Pydantic Json type for JSON or JSONB in create_pydantic_model


heliumbrain:
Example table:
```python
class Recipe(Table):
    name: str = Varchar(length=100)
    content: str = Text()
    author: BaseUser = ForeignKey(references=BaseUser)
    category: str = Varchar(length=50)
    ingredients: List = JSONB()
    extras: Dict = JSON()
```

Currently, any JSON and JSONB fields from a Piccolo Table is being serialized as strings. It technically works, but parsing it as a string outputs this:

![Screenshot 2021-06-27 at 15 02 44](https://user-images.githubusercontent.com/1607668/123545539-cd83e880-d758-11eb-896e-d53f22fa3d73.png)

Setting `value_type = pydantic.Json` for JSON and JSONB fields makes it look a lot better:

![Screenshot 2021-06-27 at 15 01 56](https://user-images.githubusercontent.com/1607668/123545610-310e1600-d759-11eb-8b7c-6cd713448954.png)

heliumbrain:
Added a PR for this here: #49 - Not sure how to link them properly in Github.. Only maintainers can do that maybe? 😕

heliumbrain:
PR merged, closing this

PR #49, opened 2021-06-27
Added support for pydantic.Json field type in create_pydantic_model

heliumbrain:
Pull request for issue #48

### DIFF
--- a/piccolo_api/crud/serializers.py
+++ b/piccolo_api/crud/serializers.py
@@ -83,7 +83,7 @@ def create_pydantic_model(
             value_type = pydantic.constr(max_length=column.length)
         elif isinstance(column, Array):
             value_type = t.List[column.base_column.value_type]  # type: ignore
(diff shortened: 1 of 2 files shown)


heliumbrain:
@dantownsend - please let me know if you want future Issues and/or pull requests structured differently. Haven't really worked in Github before, and it's been 10+ years since I did any proper development work :) Just trying to help out where I can and learn along the way.

dantownsend:
@heliumbrain Great, thanks for this.

I agree that serialising it as a proper object makes sense.  I was testing it with Piccolo admin, and Piccolo admin will need a couple of tweaks to work with this change (having an object instead of a string confuses it in places).

dantownsend:
I've made this new behaviour optional for now, until Piccolo Admin can be modified to be compatible with it.

If you're happy with it, I'll release it.

heliumbrain:
Happy! Looks like a good solution for now 👍

PR #53, opened 2021-07-11
Custom title for Swagger UI

heliumbrain:
Added the option to set a custom title for Swagger UI, like so:

```python
api.mount(
    "/docs/",
    swagger_ui(swagger_ui_title="My API docs", schema_url="../openapi.json"),
)
```

Defaults to "Piccolo Swagger UI"


dantownsend:
Cool, looks good 👍

PR #57, opened 2021-07-19
add-scripts-folder

### LINKED ISSUE #56: Add scripts folder
Following [this](https://github.com/piccolo-orm/piccolo/pull/133):

- Move all the scripts to the `scripts` folder
- Update github workflows to use the scripts

aminalaee:
Closes #56 
- Removed coverage-report.sh and added coverage report and coverage validation in test-postgres.sh and test-sqlite.sh. Looks like different coverages exist for two databases.
- Added the release.yaml, you need to get PyPI API token and set up `PYPI_TOKEN` secret for the project or organization in the Settings -> Secret.
- If you don't mind I added codecov, instead of Coveralls The results are not uploading to codecov yet. You need to set `CODECOV_TOKEN` secret for the project or organization
- Added isort (with black profile) and black to lint check to keep the codebase consistent.
- Added .flake8 and pyprojcet.toml configs to configure black and flake8. Removed # noqa: F401 from all `__init__` files and set flake8 config to ignore that in `__init__` files. But I guess a more pythonic way would be to add `__all__`
to `__init__`s so variables are not unused anymore.


scripts/lint.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+SOURCES="piccolo_api tests"
+
+isort $SOURCES
+black $SOURCES
+flake8 $SOURCES
+# mypy $SOURCES
```
dantownsend, line 8: Should we uncomment this?

dantownsend: Ah, realised there are quite a few MyPy errors at the moment, so best to keep this commented.

dantownsend:
@aminalaee Thanks for this

PR #60, opened 2021-08-06
add requirements folder and minor tweaks

sinisaos:
I really like how the `scripts` and `requirements` (thanks to @aminalaee) folders are structured in Piccolo. I think all Piccolo's repo should have the same structure.


dantownsend:
@sinisaos Looks good, thanks for this.

PR #61, opened 2021-08-07
contributing docs

sinisaos:
Adding contributing docs


dantownsend:
@sinisaos This is great - thanks!

Issue #63, opened 2021-08-09
Increase code coverage


dantownsend:
Now we have codecov integrated, it's easier to track code coverage.

There are some areas of the code base which require additional testing.

aminalaee:
@dantownsend Can this be closed?

dantownsend:
@aminalaee Yes, I can close it.

Looking at the coverage now, it's much better - thanks for that.

Only minor bits of coverage are missing now, and I can create separate issues for those.

PR #66, opened 2021-08-10
uplift-test-coverage

aminalaee:
Fixing #63 

Increases endpoint tests to 85% minimum.


piccolo_api/jwt_auth/middleware.py
```diff
@@ -96,9 +82,9 @@ async def __call__(self, scope, receive, send):
         if await self.blacklist.in_blacklist(token):
             raise HTTPException(status_code=403, detail="Token revoked")
 
-        token_dict = jwt.decode(token, self.secret)
-
-        if self.has_expired(token_dict):
+        try:
+            token_dict = jwt.decode(token, self.secret, algorithms=["HS256"])
```
aminalaee, line 86: I think it's safe to remove `expiry` from the JWT payload, since that's already done by JWT and will be handled by that,

dantownsend: You're right, good catch. I just checked the docs:

https://pyjwt.readthedocs.io/en/latest/usage.html#expiration-time-claim-exp

sinisaos:
@aminalaee  I had to change [this line]( https://github.com/piccolo-orm/piccolo_api/blob/bc8c3ef96fdda1315f9f594e0c817c0d7e533c6f/tests/fastapi/test_fastapi_endpoints.py#L56) to         `Movie(id=1, name="Star Wars", rating=93).save().run_sync()` to pass postgres tests. BTW this is great :)

piccolo_api/jwt_auth/endpoints.py
```diff
@@ -37,21 +32,17 @@ async def post(self, request: Request) -> JSONResponse:
         if not user_id:
             raise HTTPException(status_code=401, detail="Login failed")
 
-        expiry = datetime.now() + self._expiry
-
-        payload = jwt.encode({"user_id": user_id, "exp": expiry}, self._secret)
```
dantownsend, line 42: We might have keep the `exp` value to the payload. I think this is how PyJWT works out if it has expired or not.

https://github.com/jpadilla/pyjwt/blob/5fe7f2b28ffcd27a3e520be750858944889112ef/jwt/api_jwt.py#L144-L145

aminalaee: @dantownsend I think those are the claim names JWT generates itself. not our expiry in payload.
https://pyjwt.readthedocs.io/en/stable/usage.html#registered-claim-names

dantownsend: @aminalaee  I can't see any other way of setting the expiry for the token. I don't think PyJWT sets it for us - at least I couldn't see anything in the codebase, but could be mistaken.

aminalaee: @dantownsend You're right. It's the next paragraph in the docs. I'll restore that.

dantownsend:
@aminalaee This is looking great. I'm happy to merge it, unless you think anything else needs adding.

aminalaee:
@dantownsend I agree. It can be merged.

PR #67, opened 2021-08-10
Add option to exclude columns when creating pydantic model

kucera-lukas:
Simple change which adds `exclude_columns` argument to the `pydantic` creator. The specified `exclude_columns` are then skipped in the main `for` loop.

### DIFF
--- a/piccolo_api/crud/serializers.py
+++ b/piccolo_api/crud/serializers.py
@@ -39,6 +39,7 @@ def pydantic_json_validator(cls, value):
 @lru_cache()
 def create_pydantic_model(
     table: t.Type[Table],
+    exclude_columns: t.Set[str] = set(),
     include_default_columns: bool = False,
     include_readable: bool = False,
     all_optional: bool = False,
@@ -51,6 +52,9 @@ def create_pydantic_model(
     :param table:
         The Piccolo ``Table`` you want to create a Pydantic serialiser model
         for.
+    :param exclude_columns:
+        ``Set`` containing names of columns that should be exluced
+        from the Pydantic model.
     :param include_default_columns:
         Whether to include columns like ``id`` in the serialiser. You will
         typically include these columns in GET requests, but don't require
@@ -79,9 +83,15 @@ def create_pydantic_model(
         if include_default_columns
         else table._meta.non_default_columns
     )
+    
+    if not all(column in table._meta.columns for column in exclude_columns):
+        raise ValueError(f"Exclude columns ({exclude_columns!r}) are invalid.")
 
     for column in piccolo_columns:
         column_name = column._meta.name
+        if column_name in exclude_columns:
+            continue
+          
         is_optional = True if all_optional else not column._meta.required
 
         #######################################################################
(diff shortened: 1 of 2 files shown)


dantownsend:
@kucera-lukas Thanks for this, I think it's a good idea.

kucera-lukas:
Great! I'm gonna add some tests to make sure it works as expected.

piccolo_api/crud/serializers.py
```diff
@@ -39,6 +39,7 @@ def pydantic_json_validator(cls, value):
 @lru_cache()
 def create_pydantic_model(
     table: t.Type[Table],
+    exclude_columns: t.Set[str] = set(),
```
dantownsend, line 42: One thing to be aware of is this will have to be a tuple instead.

It's because the function is wrapped in `lru_cache` it won't allow any mutable types.

You might consider passing in Piccolo columns rather than strings too, so something like this:

```python
exclude_columns: t.Tuple[Column, ...] = tuple(),
```

piccolo_api/crud/serializers.py
```diff
@@ -51,6 +52,9 @@ def create_pydantic_model(
     :param table:
         The Piccolo ``Table`` you want to create a Pydantic serialiser model
         for.
+    :param exclude_columns:
+        ``Set`` containing names of columns that should be exluced
```
dantownsend, line 56: Small typo here - `excluded`

piccolo_api/crud/serializers.py
kucera-lukas: Amazing, I'm so used to working with strings. This is much cleaner

piccolo_api/crud/serializers.py
```diff
@@ -80,8 +85,22 @@ def create_pydantic_model(
...
+        # make sure that every ``column`` is tied to the current ``Table``
+        and column._meta.table is table
+        for column in exclude_columns
+    ):
+        raise ValueError(f"Exclude columns ({exclude_columns!r}) are invalid.")
+
     for column in piccolo_columns:
+        # normal __contains__ checks __eq__ as well which returns ``Where``
+        # instance which always evaluates to ``True``
+        if any(column is obj for obj in exclude_columns):
```
aminalaee, line 99: Can this be changed to `if column in exclude_columns:`? I guess it's simpler.

aminalaee:
@kucera-lukas I haven't actually checked but something is missing coverage, I guess it might be one of the `if` branchings.

kucera-lukas:
> @kucera-lukas I haven't actually checked but something is missing coverage, I guess it might be one of the `if` branchings.

Hi, I've just checked my `htmlcov` files and all of the new changes are tested.

There are 2 lines missing in `serializers.py` - `117` and `154`. `Text` and `Array` type columns are not tested but it should be easy to add. Is this PR suitable for it though?

aminalaee:
@kucera-lukas That's strange, it should've shown only your changes. But that's alright, I already added those tests in #66. So I don't think you need to.

dantownsend:
@kucera-lukas Thanks for this - I'll finish reviewing it tomorrow.

dantownsend:
@kucera-lukas Fantastic work on this - thanks!

PR #68, opened 2021-08-10
Correctly add default values to created pydantic models

kucera-lukas:
Right now default values are not being added (it's either `ellipsis` or `None` if we pass in `all_optional` as `True`). This is an issue because if people depend on the created `pydantic` schemas in their api endpoints the default values will not be parsed and will not be shown in `openapi` spec (eg. `pydantic` and `fastapi` schema generation).

### DIFF
--- a/piccolo_api/crud/serializers.py
+++ b/piccolo_api/crud/serializers.py
@@ -112,7 +112,7 @@ def create_pydantic_model(
         #######################################################################
 
         params: t.Dict[str, t.Any] = {
-            "default": None if is_optional else ...,
+            "default": None if is_optional else column._meta.params.get("default") or ...,
             "nullable": column._meta.null,
         }
 


dantownsend:
This is a good idea.

The only potential issue is sometimes the default is a callable.

I can see that Pydantic's ``Field`` can either accept ``default`` or ``default_factory``.

Perhaps this would work?

```python
params: t.Dict[str, t.Any] = {
    "default": None if is_optional else ...,
    "default_factory": column.get_default_value,
    "nullable": column._meta.null,
}
```

I need to dig into Pydantic some more to work out what the right approach would be.

kucera-lukas:
Hey, that definitely makes sense. Right now I'm on a vacation but definitely gonna look into this when I get back home.

dantownsend:
@kucera-lukas Cool, no rush - enjoy your vacation!

kucera-lukas:
@dantownsend Hi, I'm not sure if we can make this work because as I've discovered, every piccolo `Column` has a default value, even if user does not specify one. For example the `Integer` column has a default value `0`. So when creating the `pydantic` model we have no way of finding out whether the default has been created by the user or the library. I can't really think of any easy workaround.

Also, `pydantic` does not allow to pass both `default` and `default_factory` into the `Field`.

dantownsend:
@kucera-lukas Thanks for looking into this.

You're right, each column always has a default.

There's another attribute called `required`:

```python
class MyTable(Table):
    name = Varchar(required=True)
```

Piccolo itself does nothing with this attribute - it's just used to indicate to other tooling that the user should provide this value (Piccolo admin uses it for example).

Is this a potential solution? So we only set the default attribute on the Pydantic model if `required = False`?

dantownsend:
Thanks a lot for this - will review and merge asap.

piccolo_api/crud/serializers.py
```diff
@@ -131,9 +135,10 @@ def create_pydantic_model(
         #######################################################################
 
         params: t.Dict[str, t.Any] = {
-            "default": None if is_optional else ...,
             "nullable": column._meta.null,
         }
+        if not column._meta.required:
+            params.update({"default_factory": column.get_default_value})
```
dantownsend, line 141: It's taking me a while to get this merged in - I just need to make sure this still works with the Piccolo admin.

In the old code we're setting the default to `None` if `is_optional`.

Issue #69, opened 2021-08-12
Add a unit test for session logout


dantownsend:
The code coverage was increased dramatically in this issue: https://github.com/piccolo-orm/piccolo_api/issues/63

One missing area is a test for the session logout endpoint - would be nice to add this at some point.

dantownsend:
Fixed by https://github.com/piccolo-orm/piccolo_api/pull/70

PR #70, opened 2021-08-12
add session logout test

sinisaos:
Based on #69


dantownsend:
@sinisaos Looks great, thanks for this

PR #77, opened 2021-08-29
Add `nested` option to `create_pydantic_model`

### LINKED ISSUE #76: Support nested model created with `create_pydantic_model`
As suggested here:

https://github.com/piccolo-orm/piccolo/issues/202

It would be great if `create_pydantic_model` could generate nested models.

```python
create_pydantic_model(Band, nested_models=[Band.manager])
```

This would output something equivalent to this:

```python
from pydantic import BaseModel

class ManagerModel:
    name: str    

class BandModel:
    name: str
    manager: ManagerModel

```

aminalaee:
Fixes #76

### DIFF
--- a/piccolo_api/crud/serializers.py
+++ b/piccolo_api/crud/serializers.py
@@ -146,7 +146,8 @@ def create_pydantic_model(
         if isinstance(column, ForeignKey):
             if nested:
                 _type = create_pydantic_model(
-                    table=column._foreign_key_meta.resolved_references
+                    table=column._foreign_key_meta.resolved_references,
+                    nested=True,
                 )
 
             tablename = (
--- a/piccolo_api/crud/serializers.py
+++ b/piccolo_api/crud/serializers.py
@@ -40,6 +40,7 @@ def pydantic_json_validator(cls, value):
 @lru_cache()
 def create_pydantic_model(
     table: t.Type[Table],
+    nested: bool = False,
     exclude_columns: t.Tuple[Column, ...] = (),
     include_default_columns: bool = False,
     include_readable: bool = False,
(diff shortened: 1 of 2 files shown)


piccolo_api/crud/serializers.py
```diff
@@ -141,6 +144,11 @@ def create_pydantic_model(
         }
 
         if isinstance(column, ForeignKey):
+            if nested:
+                _type = create_pydantic_model(
+                    table=column._foreign_key_meta.resolved_references
```
aminalaee, line 149: I guess this should cal with `nested=True`

dantownsend: Yeah, that's right.

dantownsend:
@aminalaee Thanks for this - I think it's almost there, just need the nesting to be recursive.

tests/crud/test_serializers.py
```diff
@@ -256,3 +256,29 @@ class Computer2(Table):
...
+            name = Varchar(length=10)
+            country = ForeignKey(Country)
+
+        class Movie(Table):
+            name = Varchar(length=10)
+            director = ForeignKey(Director)
+
+        MovideModel = create_pydantic_model(table=Movie, nested=True)
+        DirectorModel = create_pydantic_model(table=Director, nested=True)
+        CountryModel = create_pydantic_model(table=Country, nested=True)
```
aminalaee, line 276: I think technically we shouldn't need to pass `nested=True` for this one. But not doing so, will fail the latest assertion.
Even though `Country` doesn't have `ForeignKey` but there's a difference between `CountryModel` created with and without `nested=True` and I guess that's related to the `@lru_cache` hashing.

dantownsend: You're right, it's probably the `lru_cache`.

dantownsend:
@aminalaee Thanks for adding this, I'm very excited about this feature.

Issue #82, opened 2021-09-07
`session_logout` should render a logout button when receiving a GET request


dantownsend:
Based on this discussion:

https://github.com/piccolo-orm/piccolo_api/issues/81

The `session_logout` endpoint only works when you send a POST request.

We can modify this endpoint to return a simple logout button when a GET request is received (in a similar way to how `session_login` works). It will make `session_logout` easier to use out of the box.

PR #84, opened 2021-09-07
session_logout to use out of the box

sinisaos:
Based on #82

### DIFF
--- a/piccolo_api/session_auth/endpoints.py
+++ b/piccolo_api/session_auth/endpoints.py
@@ -30,6 +30,10 @@ LOGIN_TEMPLATE_PATH = os.path.join(
     os.path.dirname(__file__), "templates", "login.html"
 )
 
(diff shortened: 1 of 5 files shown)


dantownsend:
Thanks for this. I added some extra content to the logout template, as on reflection just having a button was a bit plain.

I think it's pretty much good to go. We just need to update the docs.

It might be possible to move the repeated bits of the template into a base template.

sinisaos:
@dantownsend Sorry for this. I force pushed commit and delete your commits. Can you push your commits with docs and test again. Sorry for this

dantownsend:
@sinisaos No problem.

sinisaos:
@dantownsend Thanks a lot and sorry again. I added `base.html` and I think that should be it.

dantownsend:
@sinisaos No problem - it was equally my fault. I've been able to merge the changes in.

dantownsend:
Thanks for adding this.

Issue #83, opened 2021-09-07
Add a complete session auth example to the docs


dantownsend:
It would be useful to show an app which has all of the session auth components working together (`session_login`, `session_logout` and `SessionsAuthBackend`).

Here's an example:

```python
import datetime
from fastapi import FastAPI
from piccolo_api.csrf.middleware import CSRFMiddleware
from piccolo_api.openapi.endpoints import swagger_ui
from piccolo_api.session_auth.endpoints import session_login, session_logout
from piccolo_api.session_auth.middleware import SessionsAuthBackend
from starlette.middleware import Middleware
from starlette.middleware.authentication import AuthenticationMiddleware
from starlette.routing import Route

app = FastAPI()

app.mount(
    "/login/",
    session_login(),
)

private_app = FastAPI(
    routes=[
        Route("/logout/", session_logout()),
    ],
    middleware=[
        Middleware(
            AuthenticationMiddleware,
            backend=SessionsAuthBackend(
                increase_expiry=datetime.timedelta(minutes=30)
            ),
        ),
        Middleware(CSRFMiddleware, allow_form_param=True),
    ],
    docs_url=None,
    redoc_url=None,
)

# The Swagger docs which come with FastAPI don't support CSRF middleware, so we mount
# a custom one which Piccolo provides (accessible at /private/docs):
private_app.mount("/docs/", swagger_ui(schema_url="/private/openapi.json"))

@private_app.get('/my-secret-endpoint/')
def my_endpoint():
    # This is just a normal FastAPI endpoint, and is protected by Session Auth
    pass

app.mount("/private/", private_app)

if __name__ == "__main__":
    import uvicorn
    uvicorn.run(app)
```

sinisaos:
@dantownsend Do you think of the whole app as in the [Piccolo examples](https://github.com/piccolo-orm/piccolo_examples)? If not your example from comment is perfect example of `SessionAuth` and that should be in docs.

dantownsend:
@sinisaos The example code here might be sufficient.

I was thinking of a new page in the docs, with this code in it.

<img width="570" alt="Screenshot 2021-09-07 at 13 59 21" src="https://user-images.githubusercontent.com/350976/132348953-27ae6940-9db2-485e-8ad3-d3ef4ccebe8c.png">

The example code is an abbreviated form of one of my own apps - I haven't actually tested it standalone, but it looks correct.

[...]

PR #86, opened 2021-09-07
a full example of session auth

sinisaos:
Based on #83

### DIFF
--- a/docs/source/session_auth/example.rst
+++ b/docs/source/session_auth/example.rst
@@ -1,5 +1,20 @@
-Full example
+Full Example
 ============
+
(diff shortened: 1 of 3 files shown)


dantownsend:
@sinisaos Thanks for this. I managed to simplify the example slightly.

Something I've been looking for forever is a way to embed multiple files in the docs. So lets say a demo app contains 3 different files, there's a sidebar with those files in it.

Screenshot 2021-09-08 at 16 05 33

If you ever come across such a thing let me know!

For now though, I think we have a good enough solution being able to embed a single file.

Issue #87, opened 2021-09-08
Add CSRF example


dantownsend:
Similar to what we did for Session Auth, maybe we can add a full example app?

sinisaos:
@dantownsend I don’t think we need this because the addition of `CSRFMiddleware` is very clearly explained in the docs. To use `CSRFMiddleware` we need only this.
```python
from fastapi import FastAPI
from piccolo_api.csrf.middleware import CSRFMiddleware

my_asgi_app = FastAPI()

app = CSRFMiddleware(my_asgi_app, allowed_hosts=["foo.com"])
```
or
```python
from fastapi import FastAPI
from starlette.middleware import Middleware
from piccolo_api.csrf.middleware import CSRFMiddleware

app = FastAPI(
    routes=[...],
    middleware=[
        # CSRF middleware provides additional protection for older browsers, as
        # we're using cookies.
        Middleware(CSRFMiddleware, allow_form_param=True),
   ],
)
```
Maybe you can put link to [SessionAuth full example](https://piccolo-api.readthedocs.io/en/latest/session_auth/example.html) to `CSRFMiddleware` docs because that example also use `CSRFMiddleware`

dantownsend:
I guess we could add an example of a HTML template, which adds the token as form value. Or an AJAX call which adds it to the headers. I think that's the bit which is missing at the moment.

sinisaos:
@dantownsend I didn't know you thought so. For form example you can ad this
```html
<form method="POST">
    <label>Username</label>
    <input type="text" name="username" />
    <label>Password</label>
    <input type="password" name="password" />

    {% if csrftoken and csrf_cookie_name %}
    <input type="hidden" name="{{ csrf_cookie_name }}" value="{{ csrftoken }}" />
    {% endif %}

    <!-- This tells the endpoint to returns a HTML reponse if login fails. -->
    <input type="hidden" value="html" name="format" />

    <button>Login</button>
</form>
```
and for AJAX
```html
<form method="POST">
    <label>Username</label>
    <input type="text" name="username" />
    <label>Password</label>
    <input type="password" name="password" />

    {% if csrftoken and csrf_cookie_name %}
    <input type="hidden" name="{{ csrf_cookie_name }}" value="{{ csrftoken }}" />
    {% endif %}

    <!-- This tells the endpoint to returns a HTML reponse if login fails. -->
    <input type="hidden" value="html" name="format" />

    <button>Login</button>
</form>
<script>
[...]

PR #88, opened 2021-09-16
add CSRF examples

sinisaos:
Based on #87


dantownsend:
@sinisaos Thanks for this

Issue #89, opened 2021-09-19
SessionsAuthBackend.authenticate honor allow_unauthenticated if not token


Bakz:
Hello, sorry if this is a naive assertion, but shouldn't SessionsAuthBackend.authenticate return (AuthCredentials(scopes=[]), UnauthenticatedUser()) if not token but self.allow_unauthenticated == True?

dantownsend:
@Bakz you're correct - if `allow_unauthenticated` is `True`, and the middleware can't find a session for that user, it should return `(AuthCredentials(scopes=[]), UnauthenticatedUser())`.

See here:

https://github.com/piccolo-orm/piccolo_api/blob/6273f7b72204f7514d3ac566cd7980b25405c81d/piccolo_api/session_auth/middleware.py#L94

Does that help?

dantownsend:
Ah, I see you point now. It's here:

https://github.com/piccolo-orm/piccolo_api/blob/6273f7b72204f7514d3ac566cd7980b25405c81d/piccolo_api/session_auth/middleware.py#L75

You're correct - this looks like a bug, where it will deny them if they don't have a cookie.

dantownsend:
@Bakz I've released the fix - it's v0.25.1 on PyPI.

If it fixes your issue, I'll close this. Thanks.

Bakz:
@dantownsend saw that you also added the if not user_id case - all looks good now. Btw heard about you and your work on Talk Python. I'm loving piccolo - thank you for all your hard work!

dantownsend:
@Bakz Thanks for the kind words. If you find any other issues, or have any suggestions for improvements let me know!

dantownsend:
Fixed by https://github.com/piccolo-orm/piccolo_api/pull/90

PR #90, opened 2021-09-19
if `allow_unauthenticated=True`, don't reject if missing cookie

dantownsend:
Related to https://github.com/piccolo-orm/piccolo_api/issues/89

`SessionAuthMiddleware` has an option called `allow_unauthenticated`. Rather than rejecting users if they don't have a valid session, it adds an `UnauthenticatedUser` user to the request, and it's up to the downstream code to handle this.

It's a useful feature, because sometimes you want a page which supports authenticated and non-authenticated users (for example, on a website's homepage).

There was a bug where it was a too aggressive about rejecting users when `allow_unauthenticated=True`.


sinisaos (commented):
@dantownsend  That looks great.

dantownsend:
@sinisaos Cool, thanks!

Issue #79, opened 2021-09-01
Overhaul Pydantic docs


dantownsend:
Piccolo API has great Pydantic support, but the docs are a bit hidden away under the 'CRUD' section.

It would be good to move them, so 'Serialisation' is a top level topic in the docs, and contains much more examples about how to use `create_pydantic_model`.

sinisaos:
@dantownsend I'll make PR for Piccolo and Piccolo API to change docs.

sinisaos:
@dantownsend Can you please close this issue because it is done in main `Piccolo ORM` repo.

dantownsend:
@sinisaos Yep - good point

PR #95, opened 2021-10-20
updated serialization docs

sinisaos:
Related to #79


sinisaos:
@dantownsend I had to update the session auth tests again to pass the PR tests. Can you check that these updated tests are working properly locally with you. If yes, you can merge these tests to master with this PR so we have no more problems with that. Thanks.

Issue #97, opened 2021-10-24
Be able to request a subset of columns from PiccoloCRUD's GET endpoint


dantownsend:
For example:

```
/api/tables/movie?__fields=id,name
```

Would return:

```json
{
    "rows": [
        {
            "id": 17,
            "name": "The Hobbit: The Battle of the Five Armies",
        },
     ]
}
```

This is generally useful, but would be great for Piccolo Admin too, as when a user uses a ``TableConfig`` to specify that the list page should only show a subset of columns, we aren't over fetching data.

dantownsend:
@wmshort This feature is pretty much done now, except I'm not sure on the naming of the parameter.

What do you think is most suitable? `__fields`, `__columns`, `__visible_fields`?

wmshort:
Personally I find 'fields' the most intuitive but any of these will be clear enough. Glad this has come together. Nice!

dantownsend:
@wmshort Thanks - naming stuff is hard! It's good to have a sanity check.

sinisaos:
@dantownsend Can you please close this issue because it's done with latest changes.

dantownsend:
Fixed by https://github.com/piccolo-orm/piccolo_api/pull/98

PR #98, opened 2021-10-27
added ability to request a subset of columns

sinisaos:
Related to #97


sinisaos:
@dantownsend Can you please review this? Thanks.

piccolo_api/crud/endpoints.py
```diff
@@ -446,6 +455,11 @@ def _split_params(params: t.Dict[str, t.Any]) -> Params:
                     response.page_size = page_size
                 continue
 
+            if key == "__visible_fields":
+                visible_fields = str(value)
```
dantownsend, line 459: This might not be needed - is it already a string?

piccolo_api/crud/endpoints.py
```diff
@@ -529,6 +543,27 @@ async def get_all(
...
+                    [i._meta.name for i in self.table._meta.columns],
+                    self.table._meta.columns,
+                )
+            }
+
+            columns_visible: t.List[Column] = []
+            columns_non_visible: t.List[Column] = []
+
+            for key, value in all_columns_map.items():
+                if key in visible_fields.split(","):
```
dantownsend, line 560: It would be more efficient to move `visible_fields.split` out of the for loop.

piccolo_api/crud/endpoints.py
```diff
@@ -529,6 +543,27 @@ async def get_all(
...
 
+        visible_fields = split_params.visible_fields
+        if visible_fields:
+            all_columns_map: t.Dict[str, Column] = {
+                key: value
+                for key, value in zip(
+                    [i._meta.name for i in self.table._meta.columns],
+                    self.table._meta.columns,
+                )
+            }
```
dantownsend, line 554: Can you do this instead?

```python
all_columns_map = {
    i._meta.name: i for i in self.table_meta.columns
}
```

piccolo_api/crud/endpoints.py
```diff
@@ -199,14 +200,19 @@ def pydantic_model_optional(self) -> t.Type[pydantic.BaseModel]:
...
+    ):
         """
         This is for when we want to serialise many copies of the model.
         """
-        base_model = create_pydantic_model(
+        base_model: t.Any = create_pydantic_model(
             self.table,
             include_default_columns=True,
             include_readable=include_readable,
+            exclude_columns=exclude_columns,
```
dantownsend, line 215: This makes me realise that `create_pydantic_model` should probably have an `include_columns` argument.

dantownsend:
@sinisaos It looks good, thanks. Have left some comments.

piccolo_api/crud/endpoints.py
sinisaos: Yes it's a string. I'll change this

piccolo_api/crud/endpoints.py
sinisaos: You are right. I'll change this to
```python
visible_fields_split: t.List[str] = visible_fields.split(",")

for key, value in all_columns_map.items():
    if key in visible_fields_split:
        columns_visible.append(value)
    else:
        columns_non_visible.append(value)
```

piccolo_api/crud/endpoints.py
sinisaos: You're right. It’s a lot simpler and better.

piccolo_api/crud/endpoints.py
sinisaos: That's a good idea. This could be useful for some things.

sinisaos:
@dantownsend If you are OK with my replies, I can make these changes.

dantownsend:
@sinisaos Yeah, they make sense.

dantownsend:
@sinisaos Thanks for updating it to use `include_columns`. `requirements.txt` just needs updating to make sure `piccolo>=0.59.0`.

I wonder if it's possible to make it work with FastAPI? We would need to modify this:

https://github.com/piccolo-orm/piccolo_api/blob/dc139c710e23ae8a01bb47c9b267a00245596969/piccolo_api/fastapi/endpoints.py#L446

sinisaos:
@dantownsend It is already there and it works. 
https://github.com/piccolo-orm/piccolo_api/blob/aef919694c8b402208239539da6fc3e9a66ba1b4/piccolo_api/fastapi/endpoints.py#L492

Do you want to move it here at line `447` like this?

```Python
if http_method == HTTPMethod.get:
    parameters.extend(
        [
            Parameter(
                name="__visible_fields",
                kind=Parameter.POSITIONAL_OR_KEYWORD,
                annotation=str,
                default=Query(
                    default=None,
                    description=(
                        "Visible filelds to change rows result "
                        "(default ' ')."
                    ),
                ),
            ),
        ]
    )
```

dantownsend:
@sinisaos My mistake - didn't see that! Good to know it works. I'll try and get this released today.

sinisaos:
@dantownsend Great. Thanks

dantownsend:
@sinisaos I haven't merged this in yet, because I wanted to see if I could get joins working. For example:

```
/tables/movie?__visible_columns=name,director.name
```

I'm getting close - would be cool if it works.

sinisaos:
@dantownsend I'm sorry, but I didn't get to review the latest PR in Piccolo ORM repo, but something isn't working [there](https://github.com/piccolo-orm/piccolo/commit/bffd3df448afd4bca123cab340f830984b42f0fb). Result of joined columns from `create_pydantic_model` (eg. post_user.username) is always `None`. I saw that unittests passed in PR, but I can't get those results either in `create_pydantic_model` or in `visible_columns` (joined columns are not displayed). Other table columns working as expected. Am I doing something wrong or?

dantownsend:
@sinisaos I have a bunch of stuff to push to this PR to make the join functionality work.

I'll push it now so you can try it if you want. The tests will fail here, because it depends on the latest code in Piccolo master, which hasn't been released yet.

dantownsend:
To make it work is something like `PiccoloCRUD(Band, max_joins=1)`

sinisaos:
@dantownsend Sorry i didn't know that. With these change everything works great both FastAPI or Starlette :)

FireShot Capture 042 - Piccolo headless blog - Swagger UI - localhost

dantownsend:
@sinisaos Yep, sorry about that confusion.

I think it's a cool feature.

The only thing I'm still considering is what the parameter name should be. What do you think of `visible_fields` vs `fields` vs `columns`? I wonder which is most intuitive / easy to remember.

sinisaos:
@dantownsend No worries. I also think this is a **very cool** feature. As you wish, any name is good. I like `visible_fields` the most because it explicitly says which fields will be displayed in the rows result.

dantownsend:
@sinisaos The only thing I'm not sure about is whether to return the data nested or not. It wouldn't be a huge change.

Taking your example, instead of this:

```json
{
  "rows": [
    {
      "id": 2,
      "title": "Post 2",
      "post_user.username": "pero"
    }
  ]
}
```

It would be this:

```json
{
  "rows": [
    {
      "id": 2,
      "title": "Post 2",
      "post_user": {
        "username": "pero"
      }
    }
  ]
}
```

The behaviour could be controlled using an extra parameter - something like `__nested`. I think I prefer the nested approach, what do you think?

sinisaos:
@dantownsend Either way is good, but nested would be nicer.

dantownsend:
@sinisaos I've updated it now, so the response is nested. I'm much happier with it that way.

Can you give it a go when you have a chance? You'll need the latest code from Piccolo for it to work.

I've put a stupid amount of thought into what the parameter name should be. What do you think of `__response_fields`? It's either that or `__visible_fields`.

sinisaos:
@dantownsend Code looks great to me, but I want to try it out. Unfortunately, I can't try this new feature until tomorrow. When I try this, I will write a comment here. I like `__visible_fields` the most, but `__response_fields` is also a good name. I leave **"joy"** of choosing a name to you :)

dantownsend:
@sinisaos There's no rush, thanks. Naming stuff is hard!

dantownsend:
@sinisaos Great thanks. I'm pleased with it too - will get it released tomorrow then.

PR #105, opened 2021-11-11
remove body from 204 response

dantownsend:
There's an issue where returning a 204 status code (which means no response) with a body can cause an error.

Hopefully it fixes https://github.com/piccolo-orm/piccolo_admin/issues/126


trondhindenes:
Can confirm that this resolves my issue at least. :+1:

Issue #11, opened 2021-01-03
Be able to filter bulk DELETE and GET queries by a list of row IDs


dantownsend:
Currently, ``PiccoloCRUD`` (and hence ``FastAPIWrapper``) allow you to filter by the row ID, but you can't filter by a list of IDs.

For example:

```
# Currently supported:
DELETE /api/tables/movies?id=1

# Proposed:
DELETE /api/tables/movies?id=1,2,3,4,5
```

This would be a very useful feature, and would make bulk deletes more efficient.

sinisaos:
@dantownsend What do you think of this idea?

PR #107, opened 2021-11-13
added bulk delete option

sinisaos:
Related to #11


sinisaos:
@dantownsend What do you think about this? Can it be useful?

dantownsend:
@sinisaos It's definitely useful. Just need to test it locally.

sinisaos:
@dantownsend Great. Thanks.

docs/source/crud/piccolo_crud.rst
```diff
@@ -14,7 +14,7 @@ Endpoints
 ========== ======================= ==========================================================================================================
 Path       Methods                 Description
 ========== ======================= ==========================================================================================================
-/          GET, POST, DELETE       Get all rows, post a new row, or delete all matching rows.
+/          GET, POST               Get all rows or post a new row.
```
dantownsend, line 17: I think we should keep DELETE here.

piccolo_api/crud/endpoints.py
```diff
@@ -612,18 +621,24 @@ async def post_single(
...
+        split_params_ids = split_params.ids.split(",")
 
         try:
             query = self._apply_filters(
-                self.table.delete(force=True), split_params
+                self.table.delete().where(
+                    self.table._meta.primary_key.is_in(
+                        [int(item) for item in split_params_ids]
+                    )
+                ),
```
dantownsend, line 640: Here we should check if there are any valid ids before applying the filter, otherwise it will fail.

```python
query = self.table.delete()
ids = [int(item) for item in split_params_ids]
if ids:
    query = query.where(self.table._meta.primary_key.is_in(ids))
query = self._apply_filters(query, split_params)
```

tests/crud/test_crud_endpoints.py
```diff
@@ -804,17 +804,19 @@ def test_no_bulk_delete(self):
 
     def test_bulk_delete(self):
         """
-        Make sure that bulk deletes are only allowed is allow_bulk_delete is
-        True.
+        Make sure that bulk deletes are only allowed is ``allow_bulk_delete``
```
dantownsend, line 807: There's an old typo here - should be `if` instead of `is`.

docs/source/crud/piccolo_crud.rst
```diff
@@ -196,6 +196,24 @@ For example, to return results 11 to 20:
...
+
+.. code-block::
+
+    DELETE https://demo1.piccolo-orm.com/api/tables/movie/?__ids=1,2,3
+
+.. warning:: To be able to provide a bulk delete action, we must set 
+ ``allow_bulk_delete`` to ``True``.
+
+-------------------------------------------------------------------------------
+
```
dantownsend, line 216: It would be good for us to expand upon these docs some more at some point. You should be able to do stuff like:

```
DELETE /movie/?__name=foobar
```

dantownsend:
@sinisaos This is a great start. I've left some comments.

piccolo_api/crud/endpoints.py
sinisaos: Yes, you're right! I changed the `delete_bulk` method to work only with filter params (e.g. name param), only with `__ids` param and to work together for additional safety. 

```python
try:
    query: t.Union[
        Select, Count, Objects, Delete
    ] = self.table.delete()
    try:
        # if __ids query param is not empty
        ids = [int(item) for item in split_params_ids]
        query_ids = query.where(
            self.table._meta.primary_key.is_in(ids)
        )
        query = self._apply_filters(query_ids, split_params)
    except ValueError:
        # if __ids query param is empty
        query = self._apply_filters(query, split_params)
  except MalformedQuery as exception:
    return Response(str(exception), status_code=400)
```
It's a little complicated but it works.

docs/source/crud/piccolo_crud.rst
sinisaos: You're right. We can write that all three methods work.

`DELETE /movie/?__name=foobar` (delete all movies with name foobar)
`DELETE /movie/?__ids=1,2,3` (delete all movies with id 1,2,3)
`DELETE /movie/?__name=foobar&__ids=1,2` (delete all movies with name foobar, but only if id is 1 or 2)

sinisaos:
@dantownsend I made proposed changes. Can you check it out?

dantownsend:
@sinisaos Cool, thanks for this.

I was thinking about this last night. What happens when we do `DELETE /?id=1&id=2`. Does this work without the `__ids` parameter?

What you've added is a good solution, but it's part of a wider problem - being able to specify multiple values for a filter.

Imagine we wanted to filter by name: `GET /?name=StarWars&name=LordOfTheRings`.

I wonder if there's some way of making this work?

If not we can do with the `__ids` solution, but just want to consider all options.

sinisaos:
> I was thinking about this last night. What happens when we do `DELETE /?id=1&id=2`. Does this work without the `__ids` parameter?

No. It doesn't work.

> Imagine we wanted to filter by name: `GET /?name=StarWars&name=LordOfTheRings`.

and this also doesn't work due to a validation error because we can't filter with the same query params as `model_dict` value for `name` is a `list`, not an single value. 
> I wonder if there's some way of making this work?

I don't see any other way except to do the same as for `__ids` but for all possible filter params.

I think `__ids` is a convenient and easy way to pass multiple values to single query param (very similar to what we did for `__visible_fields`). If you know how to do it without `__ids`, feel free to make these changes.

dantownsend:
@sinisaos I vaguely remember being able to pass in multiple values, but I might be confusing it with something else.

Was it by doing something like: `?id[]=1,2,3`?

FastAPI makes it more difficult as it expects a single integer. But maybe if the type annotation is changes to `t.List[int]` it could work.

I'll have a look tomorrow. It probably isn't doable, but just in case.

sinisaos:
Closed - open for too long without rejection, merge or activity. Feel free to reopen PR and use it again if code is still relevant.

dantownsend:
@sinisaos All of these PRs are still relevant. I'm sorry I haven't been able to merge them in yet.

What tends to happen is a bug comes along which takes higher priority, or I don't have time to finish reviewing it.

I think moving forward there needs to be a better system around code review, so good PRs don't languish.

sinisaos:
@dantownsend No worries. I know there are a lot of things that have higher priority (especially in Piccolo ORM). I closed all PRs older than 3-4 months because nothing happens in them and it makes no sense for them to be open after all code changes that have happened in the meantime. Feel free to reopen it or use it again if you find something interesting. Cheers.

Issue #109, opened 2021-11-30
No CRUD generation support for ARRAY types.


gmos:
At version 0.29.0 there seems to be no support for ARRAY types in the CRUD generation.
Code is generated, but an ARRAY of VARCHAR comes out as a single Varchar, and the field cannot be properly used.

In the Piccolo Admin the  Arrays of Varchar are handled ok.   Are there already plans to support Array types in CRUD?

dantownsend:
@gmos Thanks for reporting this.

``PiccoloCRUD`` should support arrays - there could be a bug somewhere.

In the example below, the `years_nominated` field is an array. The API should render it like a list.

<img width="457" alt="Screenshot 2021-12-02 at 10 05 43" src="https://user-images.githubusercontent.com/350976/144400846-51c6be7b-4b7e-4cf7-b88b-63c9e40eeea8.png">

Is the problem when trying to post a new value back?

sinisaos:
@gmos @dantownsend `POST` also work as expected.

![post](https://user-images.githubusercontent.com/30960668/144437430-9fad8e92-0aeb-47fc-831d-3f2833833e39.png)

dantownsend:
@sinisaos Thanks for checking that.

gmos:
Screenshots are from the swagger .../docs page.  I expanded the root GET.

Parameters docs:
![image](https://user-images.githubusercontent.com/1526544/144451030-4b712d0b-b760-4171-bb69-da1cf4bde8fe.png)
Here it goes wrong.  No array structure to be found.  The sensors field is depicted as a simple Varchar.

And this is when I make the actual GET call:
![image](https://user-images.githubusercontent.com/1526544/144452588-6bc9cc00-d2cf-4a58-8ba4-448aa05a1448.png)

As you can see the response schema is good.  the issue is about the generated parameter input for the root GET.

dantownsend:
@gmos Cool, thanks - that's really useful. We'll investigate.

gmos:
Here is the column definition from the Table subclass:
```
    sensors = Array(
        base_column=Varchar(
            length=16,
            null=False,
        )
    )
```
And here the automatic creation of the lot.
```
hoses_router = APIRouter()
hoses_crud = PiccoloCRUD(TempHoses, read_only=False)

FastAPIWrapper(
    "/hoses",
    fastapi_app=cast(FastAPI, hoses_router),  # TODO Get away with the cast
    piccolo_crud=hoses_crud,
[...]

PR #115, opened 2021-12-09
fixing openapi arrays filtering

sinisaos:
Related to #109


dantownsend:
@sinisaos Thanks a lot for this - I've tried it out locally, and seems to work well.

I like the UI you get too:

Screenshot 2021-12-09 at 10 34 14

Issue #110, opened 2021-11-30
Add a new attribute to `PiccoloCRUD`'s schema called `fields`


dantownsend:
We now have the powerful `visible_fields` filter in `PiccoloCRUD`.

It would be nice if a consumer of the API could visit the schema endpoint, and see which fields are available (this would include fields in related tables).

For example:

```python
{
    ...
    "fields": ["id", "name", "director.name", "director.id"],
    ...
}
```

So they can know which fields they can include in the GET param:

```
/?__visible_fields=name,director.name
```

dantownsend:
Fixed by https://github.com/piccolo-orm/piccolo_api/pull/114

PR #116, opened 2021-12-10
return allowed field names

dantownsend:
Building on top of https://github.com/piccolo-orm/piccolo_api/issues/110

If a user specifies an invalid field name with `__visible_fields`, the API response tells them which field names are valid.


sinisaos (approved):
This is great

dantownsend:
@sinisaos Thanks

PR #117, opened 2021-12-11
added some tests and increase coverage

sinisaos:
Some tests have been added, configuration and migration files have been omitted to increase coverage.

### DIFF
--- a/tests/session_auth/test_session.py
+++ b/tests/session_auth/test_session.py
@@ -2,6 +2,7 @@ import os
 from unittest import TestCase
 
 from piccolo.apps.user.tables import BaseUser
+from piccolo.utils.sync import run_sync
 from starlette.authentication import requires
 from starlette.endpoints import HTTPEndpoint
 from starlette.exceptions import ExceptionMiddleware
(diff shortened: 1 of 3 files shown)


tests/session_auth/test_session.py
```diff
@@ -370,3 +372,16 @@ def test_wrong_cookie_value(self):
             response.json(),
             {"is_unauthenticated_user": True, "is_authenticated": False},
         )
+
+
+class TestCleanSessions(TestCase):
+    def setUp(self):
+        SessionsBase.create_table().run_sync()
```
dantownsend, line 379: It would be nice here if we could add a session to the database table.

dantownsend:
@sinisaos Thanks a lot for this - just one minor suggestion.

tests/session_auth/test_session.py
sinisaos: @dantownsend I agree with you. I'll do it tomorrow.

dantownsend: Thanks!

dantownsend:
@sinisaos Thanks again for this

PR #120, opened 2021-12-18
hooks

trondhindenes:
As an alternative to adding hooks/signals to piccolo-orm itself, I wanted to explore if such functionality could be more easily added to piccolo-api instead. My goal is to have extensibility points in piccolo-api (and maybe thereby also in piccolo-admin) that expand on the existing crud operations.

My propsal is that `PiccoloCRUD` taks an optional `hooks` parameter, which is a list of coroutines and when to trigger them. So far I've only added `pre_save` as a proof of concept. This lets me create a test app looking like this (i've just snipped the relevant parts of the app:

```
class Customer(Table):
    email: str = Varchar(null=True, default=None)
    first_name: str = Varchar(null=True, default=None)
    last_name: str = Varchar(null=True, default=None)


class NiceCustomer(BaseModel):
    email: str = None


async def awesome_hook(row: Customer):
    dict_row = row.to_dict()
    model_thing = NiceCustomer(**dict_row)
    async with httpx.AsyncClient() as client:
        r = await client.post(
            "https://good-url.example.com",
            json=model_thing.dict()
        )


FastAPIWrapper(
    root_url="/api/v2/customer/",
    fastapi_app=app,
    piccolo_crud=PiccoloCRUD(
        page_size=50,
        table=Customer,
        read_only=False,
        hooks=[
            Hook(
                hook_type=HookType.pre_save,
                coro=awesome_hook
            )
        ]
    )
)


```

I'd be happy to continue building on this, but I want to get maintainer's initial feedback first.

### DIFF
--- a/docs/source/crud/hooks.rst
+++ b/docs/source/crud/hooks.rst
@@ -36,7 +36,7 @@ Define an async method, and register it with PiccoloCRUD:
         ]
     )
 
-You can specify multiple hooks (also per hook_type). Hooks are executed in order. Only async methods (coroutines) are currently supported.
+You can specify multiple hooks (also per hook_type). Hooks are executed in order. You can use either async or regular functions.
 
 Hook types
 ---------------
--- a/docs/source/crud/hooks.rst
+++ b/docs/source/crud/hooks.rst
@@ -30,9 +30,9 @@ Define a method, and register it with PiccoloCRUD:
 
     # Register one or multiple hooks
     app = PiccoloCRUD(table=Movie, read_only=False, hooks=[
-        Hook(hook_type=HookType.pre_save, coro=set_movie_rating_10),
-        Hook(hook_type=HookType.pre_save, coro=set_movie_rating_20),
-        Hook(hook_type=HookType.pre_delete, coro=pre_delete)
+        Hook(hook_type=HookType.pre_save, callable=set_movie_rating_10),
+        Hook(hook_type=HookType.pre_save, callable=set_movie_rating_20),
+        Hook(hook_type=HookType.pre_delete, callable=pre_delete)
         ]
     )
 
@@ -40,7 +40,7 @@ You can specify multiple hooks (also per hook_type). Hooks are executed in order
 You can use either async or regular functions.
 
 Hook types
----------------
+----------
 
 There are different hook types, and each type takes a slightly different set of inputs. 
 It's also important to return the expected data from your hook.
@@ -59,7 +59,7 @@ It takes a single parameter, ``row``, and should return the same:
 
 
     app = PiccoloCRUD(table=Movie, read_only=False, hooks=[
-        Hook(hook_type=HookType.pre_save, coro=set_movie_rating_10)
+        Hook(hook_type=HookType.pre_save, callable=set_movie_rating_10)
         ]
     )
 
@@ -79,13 +79,13 @@ Each function must return a dictionary which represent the data to be modified.
         return values
 
     app = PiccoloCRUD(table=Movie, read_only=False, hooks=[
-        Hook(hook_type=HookType.pre_patch, coro=reset_name)
+        Hook(hook_type=HookType.pre_patch, callable=reset_name)
         ]
     )
 
 
 pre_delete
-~~~~~~~~~
+~~~~~~~~~~
 
 This hook runs during DELETE requests, prior to deleting the specified row in the database.
 It takes one parameter, ``row_id`` which is the id of the row to be deleted.
@@ -97,6 +97,6 @@ pre_delete hooks should not return data
         pass
 
     app = PiccoloCRUD(table=Movie, read_only=False, hooks=[
-        Hook(hook_type=HookType.pre_delete, coro=pre_delete)
+        Hook(hook_type=HookType.pre_delete, callable=pre_delete)
         ]
     )
--- a/docs/source/crud/hooks.rst
+++ b/docs/source/crud/hooks.rst
@@ -6,7 +6,7 @@ Hooks allow executing custom code as part of processing a crud request. You can 
 Enabling a hook
 ---------------
 
-Define an async method, and register it with PiccoloCRUD:
+Define a method, and register it with PiccoloCRUD:
 
 .. code-block:: python
 
@@ -36,12 +36,14 @@ Define an async method, and register it with PiccoloCRUD:
         ]
     )
 
-You can specify multiple hooks (also per hook_type). Hooks are executed in order. You can use either async or regular functions.
+You can specify multiple hooks (also per hook_type). Hooks are executed in order. 
+You can use either async or regular functions.
 
 Hook types
 ---------------
 
-There are different hook types, and each type takes a slightly different set of inputs. It's also important to return the expected data from your hook
+There are different hook types, and each type takes a slightly different set of inputs. 
+It's also important to return the expected data from your hook.
 
 pre_save
 ~~~~~~~~
@@ -85,7 +87,7 @@ Each function must return a dictionary which represent the data to be modified.
 pre_delete
 ~~~~~~~~~
 
-This hook runs during PATCH requests, prior to changing the specified row in the database.
+This hook runs during DELETE requests, prior to deleting the specified row in the database.
 It takes one parameter, ``row_id`` which is the id of the row to be deleted.
 pre_delete hooks should not return data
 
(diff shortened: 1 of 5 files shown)


dantownsend:
@trondhindenes This makes sense to me, thanks.

Even if we were to add signals later on to Piccolo, it wouldn't invalidate this work, as you may want certain behaviour just at the API level, so the hooks would still be useful.

trondhindenes:
nice, thanks! I'll clean it up and expand the feature set a bit. Marking the PR as draft in the meantime.

trondhindenes:
@dantownsend I've made a pass at an initial implementation, it would be great if you could have a look

piccolo_api/crud/endpoints.py
```diff
@@ -196,6 +204,7 @@ def __init__(
         self.exclude_secrets = exclude_secrets
         self.validators = validators
         self.max_joins = max_joins
+        self.hooks = hooks if hooks else []
```
dantownsend, line 207: One pattern we've used in the past is to create a dictionary, so we can look up things faster, rather than doing a list comprehension each time. For example:

```python
self.hook_map = {
    group[0]: [hook for hook in group[1]]
    for group in itertools.groupby(hooks, lambda x: x.hook_type)
}
```

The lookup is then just:

```python
self.hook_map.get(HookType.pre_save)
```

Instead of:

```python
[x for x in hooks if x.hook_type == hook_type]
```

The speed boost depends on the length of the list, but is generally worth it.

piccolo_api/crud/hooks.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class HookType(Enum):
+    pre_save = "pre_save"
+    pre_patch = "pre_patch"
+    pre_delete = "pre_delete"
+
+
+class Hook:
+    def __init__(self, hook_type: HookType, coro) -> None:
```
dantownsend, line 14: We might want to rename this from `coro` to `callable`. We can then have this as the type annotation:

```python
class Hook:
    def __init__(self, hook_type: HookType, callable: t.Callable) -> None:
```

The problem with coro, is we're asking them to provide an async function, not an actual coroutine.

It also means we can support async and non-async functions. Wherever we run the hook, we can just do this:

```python
if inspect.iscoroutinefunction(hook.callable):
    await hook.callable()
else:
    hook.callable()
```

dantownsend:
@trondhindenes This looks great, thanks. I've left a couple of comments.

There's a couple of minor linter errors - mostly just line length on docstrings, and unused variables. For the typing errors, you can just ignore those using `# type: ignore` for now.

piccolo_api/crud/endpoints.py
trondhindenes: thanks! So just so I understand you're suggesting to keep the list as the "public" api, but construct the hook_map attr during init(), and use that internally?

piccolo_api/crud/hooks.py
trondhindenes: makes total sense. Will fix.

piccolo_api/crud/endpoints.py
dantownsend: Yes, that's right.

tests/crud/test_hooks.py
```diff
@@ -0,0 +1,171 @@
...
+    return row
+
+
+async def remove_spaces(row_id: int, values: dict):
+    values["name"] = values["name"].replace(" ", "")
+    return values
+
+
+async def look_up_existing(row_id: int, values: dict):
+    row = await Movie.objects().get(Movie.id == row_id).run()
```
sinisaos, line 40: Changing this line to `row = await Movie.objects().get(Movie._meta.primary_key == row_id).run()` should remove the type error.

trondhindenes: neat, thanks!

trondhindenes:
dang, noticing that it's failing on older python versions now. I only tested 3.9 locally. Will spent some more time on this.

tests/crud/test_hooks.py
```diff
@@ -0,0 +1,182 @@
...
+    rating = Integer()
+
+    @classmethod
+    def get_readable(cls):
+        return Readable(template="%s", columns=[cls.name])
+
+
+async def set_movie_rating_10(row: Movie):
+    row.rating = 10  # type: ignore
+    return row
```
sinisaos, line 24: You can use this `async def set_movie_rating_10(row: t.Any = Movie):` or `async def set_movie_rating_10(row):` to remove `# type: ignore`

tests/crud/test_hooks.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+async def set_movie_rating_20(row: Movie):
+    row.rating = 20  # type: ignore
+    return row
```
sinisaos, line 29: You can use this `async def set_movie_rating_20(row: t.Any = Movie):` or `async def set_movie_rating_20(row):` to remove `# type: ignore`

dantownsend: Another way around this is to do `row['rating'] = 20`

tests/crud/test_hooks.py
dantownsend: Another way around this is to do `row['rating'] = 10`.

trondhindenes: thanks! I just wanted to show that these objects can be used as regular piccolo objects, but we kindof do that in the documentation so it's probably not that important. I'll fix it by using dict-notation for now, feel free to comment on that!

trondhindenes:
please let me know if there's anything else I should do here!

dantownsend:
@trondhindenes Yep, I think it's basically done. The only thing I can see is some of the docs reference `coro` instead of `callable`.

docs/source/crud/hooks.rst
```diff
@@ -0,0 +1,102 @@
...
+        return values
+
+    app = PiccoloCRUD(table=Movie, read_only=False, hooks=[
+        Hook(hook_type=HookType.pre_patch, coro=reset_name)
+        ]
+    )
+
+
+pre_delete
+~~~~~~~~~
```
dantownsend, line 88: RST is super fussy - the number of characters in the underline need to match the heading. It's missing one tilde :p

docs/source/crud/hooks.rst
```diff
@@ -0,0 +1,102 @@
...
+        return row
+
+    async def pre_delete(row_id):
+        pass
+
+    # Register one or multiple hooks
+    app = PiccoloCRUD(table=Movie, read_only=False, hooks=[
+        Hook(hook_type=HookType.pre_save, coro=set_movie_rating_10),
+        Hook(hook_type=HookType.pre_save, coro=set_movie_rating_20),
+        Hook(hook_type=HookType.pre_delete, coro=pre_delete)
```
dantownsend, line 35: -> `callable`

docs/source/crud/hooks.rst
```diff
@@ -0,0 +1,102 @@
...
+
+.. code-block:: python
+
+    async def set_movie_rating_10(row: Movie):
+        row.rating = 10
+        return row
+
+
+    app = PiccoloCRUD(table=Movie, read_only=False, hooks=[
+        Hook(hook_type=HookType.pre_save, coro=set_movie_rating_10)
```
dantownsend, line 62: -> `callable`

docs/source/crud/hooks.rst
```diff
@@ -0,0 +1,102 @@
...
+.. code-block:: python
+
+    async def reset_name(row_id: int, values: dict):
+        current_db_row = await Movie.objects().get(Movie.id==row_id).run()
+        if values.get("name"):
+            values["name"] = values["name"].replace(" ", "")
+        return values
+
+    app = PiccoloCRUD(table=Movie, read_only=False, hooks=[
+        Hook(hook_type=HookType.pre_patch, coro=reset_name)
```
dantownsend, line 82: -> `callable`

trondhindenes:
> @trondhindenes Yep, I think it's basically done. The only thing I can see is some of the docs reference `coro` instead of `callable`.

whoa, fill fix those right away. Thanks!

dantownsend:
@trondhindenes Thanks for this

trondhindenes:
> @trondhindenes Thanks for this

thanks for merging! looking forward to using this in an internal project.

dantownsend:
@trondhindenes Cool, it's on PyPI now - version 0.31.0

PR #122, opened 2021-12-31
add range header option

trondhindenes:
This PR adds a few new options for range-headers. This allows the api to send data about current pages and total number of records back to the client along with the data, which can help avoid excessive calls.

adding this will make it easier to use piccolo-api as a backend for django-admin (https://marmelab.com/react-admin)

### DIFF
--- a/piccolo_api/crud/endpoints.py
+++ b/piccolo_api/crud/endpoints.py
@@ -143,6 +143,8 @@ class PiccoloCRUD(Router):
         schema_extra: t.Optional[t.Dict[str, t.Any]] = None,
         max_joins: int = 0,
         hooks: t.Optional[t.List[Hook]] = None,
+        add_range_headers: bool = False,
+        range_header_plural_name: t.Optional[str] = None,
     ) -> None:
         """
         :param table:
@@ -196,6 +198,11 @@ class PiccoloCRUD(Router):
             To see which fields can be filtered in this way, you can check
             the ``visible_fields_options`` value returned by the ``/schema``
             endpoint.
+        :param add_range_headers:
+            if True, will add content-range headers to GET responses returning lists of data
+        :param range_header_plural_name:
+            Specify object name which is included in the Content-Range header 
+            when `add_range_headers` is True. Defaults to table name if unset.
 
         """  # noqa: E501
         self.table = table
@@ -212,6 +219,8 @@ class PiccoloCRUD(Router):
             }
         else:
             self._hook_map = None  # type: ignore
+        self.add_range_headers = add_range_headers
+        self.range_header_plural_name = range_header_plural_name
 
         schema_extra = schema_extra if isinstance(schema_extra, dict) else {}
         self.visible_fields_options = get_visible_fields_options(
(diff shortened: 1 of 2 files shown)


dantownsend:
@trondhindenes Cool, thanks for this.

I didn't realise there was a range HTTP header, but it seems useful. I'll review asap.

@sinisaos What do you think?

sinisaos:
@dantownsend I also think this is great. I tried this and everything works perfect.  
@trondhindenes This is very cool :+1: .

trondhindenes:
awesome, thanks! I'll add a bit of documentation, i think some of the code can be slightly simplified aswell. Will do that in a new commit, then this one should be good to go.

trondhindenes:
should be good to go now!

docs/source/crud/piccolo_crud.rst
```diff
@@ -291,6 +291,41 @@ You can also use this on GET requests when retrieving a single row, for example:
...
+
+In some applications it can be practical to get information about the
+total number of records without invoking a separate call to
+the `count` endpoint. Piccolo API will supply this information in the
+`Content-Range` response header if `add_range_headers` is set to `True`.
+You can use the `range_header_plural_name` parameter to configure the
+"plural name" used in the `Content-Range` response header.
+
+The contents of the `Content-Range` header might look something like this
+for the "Movie" table: `movie 0-9/100`
```
dantownsend, line 305: It's a bit annoying, but can we use double backticks? RST renders single backticks as italics, and double backticks as code blocks, which look slightly nicer.

dantownsend:
This looks cool.

The only thing which confused me at first was how it's zero indexed:

```
movie 0-9/100
```

Instead of:

```
movie 1-10/100
```

But this just seems to be the way it's done.

I just made one minor comment about the docs - again, it's just RST being fussy. Other than that, seems great.

trondhindenes:
Yup, I had to read up on the indexing too :-)

Backtics are fixxed  I'll admit I didn't try rendering the docsite locally - probably should have. In any case, should be good to go now.

docs/source/crud/piccolo_crud.rst
trondhindenes: fixed!

dantownsend:
@trondhindenes Thanks, looks good

dantownsend:
One thought I just had was whether this should be a GET parameter.

So something like:

```
/?__content_header=true
```

So the user can turn it on or off dynamically. Not sure if this is a valid use case.

trondhindenes:
oh, that's a good idea. Are you thinking instead of the `Piccolocrud` init param, or as an alternative?

dantownsend:
@trondhindenes I think if we had the GET parameter we could lose the init param. 

I haven't used React Admin much, but I assume using a GET parameter instead shouldn't be an issue.

With the plural name, this could also be specified in the GET param, but not sure what you think about it:

```
/?__range_header=true&__range_header_name=movies
```

I think what we have now is really good, but the reasons I'm considering the GET param are:

 * Keeps the `PiccoloCRUD.__init__` a bit lighter, as it has a lot of options
 * The ability to turn the header on and off dynamically could be useful
 * We can add this parameter to the Swagger docs

But the downside is it's less convenient having to pass the GET parameter in for situations where you always want the range header.

@sinisaos What do you think about this?

I don't mind doing a quick refactor to use a GET param, but we just need to decide if it's a better approach or not. Once the API is released, it's there forever!

sinisaos:
@dantownsend Both approaches have pros and cons. I slightly prefer the original @trondhindenes approach as it is now, but if everyone else agrees that the new approach is better I have nothing against it. Both approaches serve a purpose.

trondhindenes:
for the react-admin use case (which was my motivation) a get param is completely acceptable, and it avoids bloating the init methods. Regarding the plural name: It's more of a convencience. I can't think of any situation where it's critical to return the plural form in the header. 

Since piccolo-api already has a fairly rich set of http parameters, it kindof fits nicely in there. I think my vote would be to refactor from init param to http param.

Want me to do it?

dantownsend:
@trondhindenes If you don't mind, that would be great.

PR #123, opened 2022-01-02
use http param instead of init parameter to control range-header behavior

trondhindenes:
as discussed in https://github.com/piccolo-orm/piccolo_api/pull/122

### DIFF
--- a/piccolo_api/fastapi/endpoints.py
+++ b/piccolo_api/fastapi/endpoints.py
@@ -515,7 +515,8 @@ class FastAPIWrapper:
                         default=Query(
                             default=False,
                             description=(
-                                "enable to add range-header respons header"
+                                "Set to 'true' to add the "
+                                "Content-Range response header"
                             ),
                         ),
                     )
@@ -530,7 +531,8 @@ class FastAPIWrapper:
                         default=Query(
                             default=None,
                             description=(
-                                "object name in range-header reponse header"
+                                "Specify the object name in the Content-Range "
+                                "response header (defaults to the table name)."
                             ),
                         ),
                     )
--- a/piccolo_api/fastapi/endpoints.py
+++ b/piccolo_api/fastapi/endpoints.py
@@ -506,6 +506,37 @@ class FastAPIWrapper:
                     ]
                 )
 
+            parameters.extend(
+                [
+                    Parameter(
+                        name="__range_header",
+                        kind=Parameter.POSITIONAL_OR_KEYWORD,
+                        annotation=bool,
+                        default=Query(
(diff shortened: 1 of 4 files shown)


sinisaos:
@trondhindenes @dantownsend You were both right. This is a much cleaner approach. Can you add a description of the `__range_header` and `__range_header_name` query params to the `_split_params` method docstring. Thanks.

trondhindenes:
> Can you add a description of the `__range_header` and `__range_header_name` query params to the `_split_params` method docstring. Thanks.

Done!

piccolo_api/fastapi/endpoints.py
```diff
@@ -506,6 +506,37 @@ def modify_signature(
...
+            parameters.extend(
+                [
+                    Parameter(
+                        name="__range_header",
+                        kind=Parameter.POSITIONAL_OR_KEYWORD,
+                        annotation=bool,
+                        default=Query(
+                            default=False,
+                            description=(
+                                "enable to add range-header respons header"
```
dantownsend, line 518: For the description, how about:

`Set to 'true' to add the Content-Range response header.`

piccolo_api/fastapi/endpoints.py
```diff
@@ -506,6 +506,37 @@ def modify_signature(
...
+            parameters.extend(
+                [
+                    Parameter(
+                        name="__range_header_name",
+                        kind=Parameter.POSITIONAL_OR_KEYWORD,
+                        annotation=str,
+                        default=Query(
+                            default=None,
+                            description=(
+                                "object name in range-header reponse header"
```
dantownsend, line 533: For the description, how about:

`Specify the object name in the Content-Range response header (defaults to the table name).`

dantownsend:
@trondhindenes This looks great, thanks!

I just left some minor suggestions for the Swagger docs descriptions.

trondhindenes:
@dantownsend thanks! Your suggestions make total sense - added those now.

dantownsend:
Cool, looks good - sorry for all the back and forth. I think this is the right solution.

trondhindenes:
@dantownsend there's a bug in this release - if `__range_header` is false it gets passed down to the model, which causes a validation error. Sorry about this, I'll create a hotfix PR right away.

trondhindenes:
https://github.com/piccolo-orm/piccolo_api/pull/125. Again, sorry for this I should have testet for that from the get-go.

dantownsend:
@trondhindenes No worries, these things happen. I'll release the hot fix when the build passes.

PR #127, opened 2022-01-03
make patch response json/dict instead of string

### LINKED ISSUE #126: PATCH endpoint behaves differently
Noticing that the response from PATH seems to be a text-type instead of json-type response. I'm noticing in the code that the PATH response uses `JSONResponse` instead of the `CustomJSONResponse` used by many other endpoints. 
As far as I can see, functions either use json response with a dict input:
```
JSONResponse(self.pydantic_model.schema()
```
or `CustomJSONResponse` with a string input:
```
        json = self.pydantic_model_plural(
            include_columns=tuple(visible_columns),
            nested=nested,
        )(rows=rows).json()
        return CustomJSONResponse(json, headers=headers)
```

the PATCH response is the only one using `JSONResponse` with a string param. 

I'm guessing this is just a bug (and not a critical one at that) but before I submit a PR to fix I just wanted to make sure.

trondhindenes:
changes patch response so that it's a proper json instead of a json-formatted string. This also means that tests no longer have to convert the string response to dict before comparing so that's a nice added bonus - a bit cleaner overall.

fixes #126

### DIFF
--- a/piccolo_api/crud/endpoints.py
+++ b/piccolo_api/crud/endpoints.py
@@ -1044,7 +1044,7 @@ class PiccoloCRUD(Router):
                 .first()
                 .run()
             )
-            return JSONResponse(self.pydantic_model(**new_row).json())
+            return CustomJSONResponse(self.pydantic_model(**new_row).json())
         except ValueError:
             return Response("Unable to save the resource.", status_code=500)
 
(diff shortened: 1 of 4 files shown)


dantownsend:
@trondhindenes Thanks, looks great 👍

PR #131, opened 2022-01-12
add PK name to schema

dantownsend:
Related to https://github.com/piccolo-orm/piccolo_admin/issues/81

Added the name of the primary key column to the schema endpoint.


sinisaos (approved):
This is great.

PR #133, opened 2022-01-21
update tests for target column

sinisaos:
Updated tests for new Piccolo ORM `target_column`.


dantownsend:
@sinisaos Thanks a lot for this 👍

PR #134, opened 2022-01-21
changes to support non-primary key in Piccolo Admin

sinisaos:
Related to [Supporting foreign keys which reference non-primary keys #144](https://github.com/piccolo-orm/piccolo_admin/pull/144)


piccolo_api/crud/endpoints.py
```diff
@@ -423,14 +423,32 @@ async def get_ids(self, request: Request) -> Response:
 
         values = await query.run()
 
+        # if value_type not in or UUID
+        try:
+            target_pk_type = [
+                i._meta.params["target_column"].value_type not in (int, uuid)
+                for i in self.table._meta._foreign_key_references
+            ][0]
```
dantownsend, line 431: I see the idea. I think the problem though is it's having to do `[0]`. If there the `_foreign_key_references` contains more than 1 item, we're guessing that the first one is correct.

I think we're going to have to add a GET parameter so the Piccolo Admin UI can explicitly tell us which column it's interested in.

sinisaos: Yes. That make sense. I will try to add a GET parameter.

dantownsend: Cool, thanks. I think another advantage of the GET param is if someone doesn't pass it we can return the same response as before, so there are no breaking API changes.

sinisaos: I didn't get far, but we can pass `target_column` as a query param 
```python
target_column = request.query_params.get("target_column")
    if target_column is not None:
        return JSONResponse({i["readable"]: i["readable"] for i in values})
```
and return the result like this
```
# ./api/tables/movie/ids/?target_column=
{
   'Star Wars': 'Star Wars',
   'Return of the Jedi: 'Return of the Jedi,
   ...
}
```
(we don't need to pass anything to query param and return all results because the `KeySearchModal` already needs all results) 
or without the `target_column` return standard response
```
# ./api/tables/movie/ids/
{
   1: 'Star Wars',
   2: 'Return of the Jedi,
   ...
}
``` 
I don't know yet how to trigger Vue frontend based on the existence of the `column_name` query parameter. I don't know is this ok, or I completely miss the point.

piccolo_api/crud/endpoints.py
```diff
@@ -871,7 +889,16 @@ async def detail(self, request: Request) -> Response:
         try:
             row_id = self.table._meta.primary_key.value_type(row_id)
         except ValueError:
-            return Response("The ID is invalid", status_code=400)
+            for i in self.table._meta._foreign_key_references:
+                target = i._meta.params["target_column"]
```
dantownsend, line 893: This should be `i._meta.params.get("target_column")` in case the key doesn't exist.

piccolo_api/crud/endpoints.py
```diff
@@ -871,7 +889,16 @@ async def detail(self, request: Request) -> Response:
...
+                if target is not None:
+                    reference_target_pk = (
+                        await self.table.select(self.table._meta.primary_key)
+                        .where(target == row_id)
+                        .first()
+                        .run()
+                    )
+                    row_id = reference_target_pk[self.table._meta.primary_key]
```
dantownsend, line 901: We can do this slightly easier:

```python
reference_target_pk = (
    await self.table.select(self.table._meta.primary_key)
    .where(target == row_id)
    .output(as_list=True)
    .first()
    .run()
)[0]
```

dantownsend:
@sinisaos I see where you're going with this. I'll have another look tomorrow at the UI side.

piccolo_api/crud/endpoints.py
```diff
@@ -871,7 +889,16 @@ async def detail(self, request: Request) -> Response:
         try:
             row_id = self.table._meta.primary_key.value_type(row_id)
         except ValueError:
-            return Response("The ID is invalid", status_code=400)
+            for i in self.table._meta._foreign_key_references:
```
dantownsend, line 892: I think we can move the logic around a bit here. 

```python
target = i._meta.params.get("target_column")

if target is None:
    try:
        row_id = self.table._meta.primary_key.value_type(row_id)
     except ValueError:
        return Response("The ID is invalid", status_code=400)
else:
    # insert your new logic here
```

piccolo_api/crud/endpoints.py
sinisaos: Unfortunately this does not work if we have multiple non-primary keys per single table.

sinisaos:
@dantownsend Basically we have a working list of rows with only primary keys, only with non-primary keys, or a mixture of both.
The problem is the filter search in sidebar and referencing tables and how to combine `id` and `readable` based on `target_colum` or `primary key`. I hope you got some ideas. This is tricky with a lot of edge cases.

piccolo_api/crud/endpoints.py
sinisaos: Unfortunately this also does not work if we have multiple non-primary keys per single table. e.g if we add director name as non-primary key (or mix primary or non-primary keys) to review reviewer this does not work. we should to leave this as is.

dantownsend:
@sinisaos It looks promising - I'll give it a go.

sinisaos:
@dantownsend We also need to make these changes to add and edit records in Piccolo Admin to make everything work.
```javascript
// in EditRowForm.vue
methods: {
    async submitForm(event) {
        console.log("Submitting...")

        const form = new FormData(event.target)

        const json = {}
        const targetColumn = []
        for (const i of form.entries()) {
            const key = i[0].split(" ").join("_")
            let value = i[1]

            if (value == "null") {
                value = null
            } else if (this.schema.properties[key].type == "array") {
                // @ts-ignore
                value = JSON.parse(value)
            } else if (
                this.schema?.properties[key].format == "date-time" &&
                value == ""
            ) {
                value = null
            } else if (
                this.schema?.properties[key].extra.foreign_key == true &&
                value == ""
            ) {
                value = null
            } else if (
                this.schema?.properties[key].extra.foreign_key == true &&
                this.schema?.properties[key].format != "uuid" &&
                this.schema?.properties[key].type != "integer" &&
                key == this.schema?.properties[key].extra.to
            ) {
                targetColumn.push(value)
            }
            json[key] = targetColumn[1] == value ? targetColumn[0] : value
        }
...
```
and
```javascript
// in AddRowForm.vue
methods: {
    async submitForm(event) {
        console.log("I was pressed")
        const form = new FormData(event.target)

        const json = {}
        const targetColumn = []
        for (const i of form.entries()) {
            const key = i[0].split(" ").join("_")
            let value = i[1]

            if (value == "null") {
                value = null
                // @ts-ignore
            } else if (this.schema.properties[key].type == "array") {
                // @ts-ignore
                value = JSON.parse(value)
                // @ts-ignore
            } else if (
                this.schema?.properties[key].format == "date-time" &&
                value == ""
            ) {
                value = null
                // @ts-ignore
            } else if (
                this.schema?.properties[key].extra.foreign_key == true &&
                value == ""
            ) {
                value = null
            } else if (
                this.schema?.properties[key].extra.foreign_key == true &&
                this.schema?.properties[key].format != "uuid" &&
                this.schema?.properties[key].type != "integer" &&
                key == this.schema?.properties[key].extra.to
            ) {
                targetColumn.push(value)
            }
            json[key] = targetColumn[1] == value ? targetColumn[0] : value
        }
...
```

sinisaos:
@dantownsend Did you have time to try this? If you agree with this, I can do PR in Piccolo Admin, with changes from the comments to make everything work.

dantownsend:
@sinisaos I haven't had a chance, but I would like to get it sorted. If you could create the PR that would be great, thanks.

sinisaos:
@dantownsend Great. I will wait with PR because these changes in Piccolo Admin are only relevant if the code from this PR is good enough.

dantownsend:
@sinisaos OK, makes sense

sinisaos:
@dantownsend Everything should work now. Combination of primary and non-primary keys per table, multiple non-primary keys per table etc.

sinisaos:
@dantownsend Have you had a chance to try this?

dantownsend:
@sinisaos Not yet - will try and review it this weekend.

sinisaos:
@dantownsend Great. Thanks.

Issue #92, opened 2021-09-27
Add hooks to `session_login`


dantownsend:
When using the `session_login` endpoint, it would be nice to have hooks (basically callback functions) which are called at different parts of the verification process.

There are many situations where this would be valuable. Here are some examples:

 1. You only want to let the user login if they have verified their email address.
 2. After the user logs in, you want to record this in an external system.
 3. Before logging a user in, you want to check if they are on a black list (for example, based on email address or IP address).

There will have to be a minimum of three hook types:

1. Triggered before a login attempt is made. 
2. Triggered after the login is successful.
3. Triggered when the login fails.

The `session_login` function will have a new `hooks` argument:

```python
def session_login(
    ...,
    hooks: t.List[SessionLoginHook] = []
):
    ...
```

The hook will accept a request argument, and additional data (such as email, `BaserUser` instance etc).

If the hook raises an `Exception`, then the login is stopped, and an error is shown to the user.

It would be great if this functionality was available for all of the auth types that we provide (token auth etc), but session auth is the most important.

PR #138, opened 2022-04-14
added `LoginHooks`

dantownsend:
Fixes https://github.com/piccolo-orm/piccolo_api/issues/92


piccolo_api/shared/auth/hooks.py
```diff
@@ -0,0 +1,136 @@
...
+
+    Here's an example using :class:`session_login <piccolo_api.session_auth.endpoints.session_login>`:
+
+    .. code-block:: python
+
+        def check_ban_list(username: str, **kwargs):
+            '''
+            An example pre_login hook.
+            '''
+            if username in ('nuisance@pest.com',):
```
sinisaos, line 41: In this example, I think you should pass tuple of usernames not emails because this always return `False` and doesn't trigger this hook. E.g `if username in ("john", "bill",):`

piccolo_api/shared/auth/hooks.py
```diff
@@ -0,0 +1,136 @@
...
+                return 'This account has been temporarily suspended'.
+
+
+        def send_email(user: BaseUser, **kwargs):
```
sinisaos, line 45: This should be `async def send_email(user: BaseUser, **kwargs):` because is a coroutine.

sinisaos (approved):
I think this is great new feature. I left a few comments (essentially typos).

piccolo_api/shared/auth/hooks.py
dantownsend: Good point - I'll change that.

piccolo_api/shared/auth/hooks.py
dantownsend: Good catch!

Issue #139, opened 2022-04-19
Add signup endpoint


dantownsend:
We have `session_login` and `session_logout` endpoints, which helps with quickly prototyping an app.

It would be great to have a `session_signup` endpoint too.

It would look very similar to the login form, except with username / email / password / confirm password fields:

<img width="575" alt="Screenshot 2022-04-19 at 19 11 03" src="https://user-images.githubusercontent.com/350976/164068599-7df4bfbc-b39b-4f95-842b-b48738ec9375.png">

sinisaos:
@dantownsend I was thinking the same thing yesterday and it would be useful. Can I try implement this? My only question is how deep should we go with email and password validation without an external library. Is the html email field enough and is the 6-character password long enough? Or should we go deeper with regular expression?

dantownsend:
@sinisaos Thanks, it would great if you could take a look. Just a simple regex should be OK for email validation. Even if the first version is very simple, it's still useful. I think we can add hooks at some point, so users can add additional logic for password and email validation.

sinisaos:
@dantownsend I will make PR during the day that you can see code changes. We will drop a little bit in coverage because multiple lines like [this](https://github.com/piccolo-orm/piccolo_api/blob/6f656a9048d482419efac329be888bdf45f8c136/piccolo_api/session_auth/endpoints.py#L182 ) one used to render errors in the HTML template. I think it's not a big deal.

dantownsend:
Implemented in https://github.com/piccolo-orm/piccolo_api/pull/141

PR #141, opened 2022-04-20
added session_signup

sinisaos:
Related to #139


dantownsend:
@sinisaos This looks good, thanks - I'll try it out tomorrow. 👍

piccolo_api/session_auth/templates/signup.html
```diff
@@ -0,0 +1,32 @@
...
+{% endif %}
+
+<form method="POST">
+    <label>Username</label>
+    <input type="text" name="username" required />
+    <label>Email</label>
+    <input type="email" name="email" required />
+    <label>Password (6 characters minimum)</label>
+    <input type="password" name="password" minlength=6 required />
+    <label>Confirm Password (6 characters minimum)</label>
```
dantownsend, line 19: I don't think we need to repeat the `(6 characters minimum)`.

dantownsend:
@sinisaos This looks good, thanks.

My main question is whether to automatically log the user in after account creation.

I found a few discussions about it:

 * https://stackoverflow.com/questions/12646520/should-user-auto-login-after-registration
 * https://ux.stackexchange.com/questions/21619/making-the-user-login-again-after-registration

We could let the programmer redirect the user to a login page after account creation using the `redirect_to` parameter.

So the app could be:

```python
app = FastAPI()

app.mount('/login', session_login())
app.mount('/sign-up', session_signup(redirect_to='/login'))
```

Because it decouples sign up from session creation, we could also call it something more generic - `sign_up` instead of `session_signup`.

What do you think?

piccolo_api/session_auth/templates/signup.html
sinisaos: You're right. I'll remove that.

sinisaos:
@dantownsend As you wish. Some websites register the user directly after registration, and some do not. I think it's much more comfortable if the user logs in right away, but you're right about decouple signup from session creation. If we decide to redirect user to login page we need only to delete some code and everything works.
If we go that way, `SignupEndpoint` is a much more appropriate name than `SessionSignupEndpoint`.

dantownsend:
@sinisaos Yeah, I think we should make it so they have to login to create a session.

I agree, it's more convenient if the sign up and login are combined, but if we keep them separate it makes the new endpoint much lighter:

```python
def signup(
    auth_table: t.Optional[t.Type[BaseUser]] = None,
    redirect_to: t.Optional[str] = "/",
    template_path: t.Optional[str] = None,
) -> t.Type[SessionSignupEndpoint]:
    ...
```

Also, when we add hooks to this new endpoint we'll be able to do things like this:

```python
app = FastAPI()

app.mount('/sign-up', signup(redirect_to='/login', hooks=SignUpHooks(on_success=[send_confirmation_email])
app.mount('/login', session_login(hooks=LoginHooks(pre_login=[check_email_confirmed]))
```

After the sign up, we can use a hook to send a confirmation email, and then when they login we can use another hook to make sure the email confirmation was used.

What do you think?

sinisaos:
@dantownsend I decoupled signup from session creation. Can you check if this is OK?

dantownsend:
@sinisaos I'm so sorry - I didn't see your last comment. I'll test it now.

dantownsend:
I moved things around slightly. I moved the new endpoint into a new folder called 'register'.

I renamed it from `signup` because it's one of those things which can be spelt in many different ways - `signup`, `sign_up` `sign-up`, and I kept confusing myself.

I think the remaining consideration is around security. The documentation definitely needs to mention that rate limiting should be used on this endpoint. Also, I'm considering if a captcha or is needed too, to prevent spam.

piccolo_api/register/endpoints.py
```diff
@@ -158,21 +164,29 @@ def register(
...
         so by specifying the absolute path to a custom template. For example
         ``'/some_directory/register.html'``. Refer to the default template at
-        ``piccolo_api/session_auth/templates/register.html`` as a basis for
+        ``piccolo_api/templates/register.html`` as a basis for
         your custom template.
+    :param user_defaults:
+        These values are assigned to the new user. An example use case is
+        setting ``active = True`` on each new user, so they can immediately
+        login (not recommended for production, as it's better to verify their
+        email address first, but OK for a prototype app)::
```
sinisaos, line 187: If we use `user_defaults` arg, we need also set a **warning** in docs that is only for prototyping because otherwise there is no point in decouple the `register` from creating the session, and we could immediately log the user in from `register` endpoint (as we did at the beginning).

dantownsend: Yeah, it's a good point. They will still have to login to get the session though.

dantownsend:
I'll add docs etc for using CAPTCHAs in a future PR.

PR #145, opened 2022-06-04
Added deletion in bulk

sinisaos:
New PR related to #11, but also added support for `UUID` primary keys.

### DIFF
--- a/docs/source/crud/piccolo_crud.rst
+++ b/docs/source/crud/piccolo_crud.rst
@@ -244,6 +244,42 @@ For example, to return results 11 to 20:
 
 -------------------------------------------------------------------------------
 
+Bulk delete
+-----------
+
+To specify which records you want to delete in bulk, pass a query parameter 
+like this ``__ids=1,2,3``, and you be able to delete all results whose ``id`` 
+is in the query params.
+
+.. hint:: You can also use this method with ``UUID`` primary keys and 
+    the usage is the same.
+
+A query which delete movies with ``id`` pass in query parameter:
+
+.. code-block::
+
+    DELETE https://demo1.piccolo-orm.com/api/tables/movie/?__ids=1,2,3
+
+You can delete rows in bulk with any filter params. A query which 
+delete movies with ``name`` pass in query parameter:
+
+.. code-block::
+
+    DELETE https://demo1.piccolo-orm.com/api/tables/movie/?name=Star
+
+Or you can combine multiple query params for additional security. 
+A query to delete records with name ``Star``, but with 
+specific ``id`` you can pass query like this: 
+
+.. code-block::
+
+    DELETE https://demo1.piccolo-orm.com/api/tables/movie/?name=Star&__ids=1,2
+
+.. warning:: To be able to provide a bulk delete action, we must set 
+ ``allow_bulk_delete`` to ``True``.
+
+-------------------------------------------------------------------------------
+
 Readable
 --------
 
--- a/piccolo_api/crud/endpoints.py
+++ b/piccolo_api/crud/endpoints.py
@@ -86,6 +86,7 @@ class Params:
     include_readable: bool = False
     page: int = 1
     page_size: t.Optional[int] = None
+    ids: str = field(default="")
     visible_fields: str = field(default="")
     range_header: bool = False
     range_header_name: str = field(default="")
@@ -155,8 +156,8 @@ class PiccoloCRUD(Router):
         :param read_only:
             If ``True``, only the GET method is allowed.
         :param allow_bulk_delete:
-            If ``True``, allows a delete request to the root to delete all
-            matching records. It is dangerous, so is disabled by default.
+            If True, allows a delete request to the root and delete all
+            matching records with values in ``__ids`` query params.
         :param page_size:
             The number of results shown on each page by default.
         :param exclude_secrets:
@@ -521,7 +522,7 @@ class PiccoloCRUD(Router):
             return await self.post_single(request, data)
         elif request.method == "DELETE":
             params = dict(request.query_params)
-            return await self.delete_all(request, params=params)
+            return await self.delete_bulk(request, params=params)
         else:
             return Response(status_code=405)
 
@@ -547,6 +548,9 @@ class PiccoloCRUD(Router):
         {'__page_size': 15}.
 
         And can specify which page: {'__page': 2}.
+
+        You can specify witch records want to delete from rows:
+        {'__ids': '1,2,3'}.
 
         You can specify which fields want to display in rows:
         {'__visible_fields': 'id,name'}.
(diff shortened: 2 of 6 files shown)


piccolo_api/crud/endpoints.py
```diff
@@ -548,6 +549,9 @@ def _split_params(params: t.Dict[str, t.Any]) -> Params:
 
         And can specify which page: {'__page': 2}.
 
+        You can specify witch records want to delete from rows:
```
dantownsend, line 552: Small typo here, 'witch -> 'which'.

piccolo_api/crud/endpoints.py
```diff
@@ -820,19 +828,31 @@ async def post_single(
...
         split_params = self._split_params(params)
+        split_params_ids = split_params.ids.split(",")
 
         try:
-            query = self._apply_filters(
-                self.table.delete(force=True), split_params
-            )
+            query: t.Union[
+                Select, Count, Objects, Delete
+            ] = self.table.delete()
```
dantownsend, line 844: Should this just be `query: Delete = self.table.delete()`

piccolo_api/crud/endpoints.py
```diff
@@ -820,19 +828,31 @@ async def post_single(
...
+            try:
+                ids = [
+                    int(item) if len(item) < len(str(uuid.uuid4())) else item
+                    for item in split_params_ids
+                ]
```
dantownsend, line 849: I would do this slightly differently - rather than working out whether it's a UUID based on the length, you can do it instead like this:

```python
isinstance(self.table._meta.primary_key, UUID) # where UUID is Piccolo's UUID column type
```

piccolo_api/crud/endpoints.py
sinisaos: Thanks. I'll change that.

piccolo_api/crud/endpoints.py
sinisaos: This will not work because `split_params_ids` is a list of strings. I can add type annotation to `split_params id` like this `split_params_ids: t.List[str] = split_params.ids.split(",")`. I added `len(str(uuid.uuid4()))` because Postgres `BigInt` is same as `sys.maxsize` which is length 19.

piccolo_api/crud/endpoints.py
sinisaos: Unfortunately that raise `mypy` error (`error: Incompatible types in assignment (expression has type "Union[Select, Count, Objects, Delete]", variable has type "Delete"`)
Because of that I added `t.Union[Select, Count, Objects, Delete]` type annotation.

piccolo_api/crud/endpoints.py
dantownsend: What about this:

```python
value_type = self.table._meta.primary_key.value_type
ids = [value_type(item) for item in split_params_ids]
```

sinisaos: This works perfectly. Thanks.

sinisaos:
@dantownsend I also added a bulk update option. With these changes we can do the proper batch actions also in Piccolo Admin (we need to change some code, it remains to be seen) and in the `audit_logs` app so we can have batch results as you mentioned. Of course there is also one problem with `uuid` primary key. As they are quite long (36 characters) we should reduce the page size in Piccolo Admin (e.g. 5,10,25,50) where 50 (or less) is the maximum number of lines because the maximum URL length is 2048. This would reduce the possibility of side effects and URL length exceeding.

dantownsend:
@sinisaos The filters are useful in bulk delete, because imagine you wanted to delete all rows where `draft=False`. You can do that with the filters. It's mostly if Piccolo API is being used standalone - I don't think we use that feature in Piccolo Admin.

The changes look good. I'm a bit behind of reviewing PRs - will try and get on top of it all next week.

sinisaos:
@dantownsend It would be great if you find time to review and merge these PRs #134, #145, #158, #160 (I think they are good and even if some edge cases occur, we can always fix it later) in the Piccolo API because after that we can make changes in the Piccolo Admin and add a lot of real nice features (I already have most of the Piccolo Admin solutions ready).

sinisaos:
@dantownsend Can you review and merge #145. After that we can make the changes in #160 and implement both of these new features also in Piccolo Admin?

dantownsend:
@sinisaos Yeah, I'll try and merge this in this week. I've started reviewing the custom image PR.

sinisaos:
@dantownsend Just a reminder. If you find time, it would be nice to merge this because after that we can update Piccolo Admin as well, and finish audit logs app.

dantownsend:
Yeah, I need to merge it in. I need to carefully review it first though, as bulk updates and bulk deletes are very powerful, and we don't want there to be any unforeseen issues.

sinisaos:
You are right that we should be careful because these are dangerous actions, but these actions have an effect on rows only if we pass primary keys to `ids` list, otherwise nothing happens.

dantownsend:
That's a good point - if it's driven by the IDs you pass in, then less chance of accidentally modifying other rows.

sinisaos:
Yes. Exactly that. I don't think there is any fear of modifying the record if we don't pass that record PK to `ids` list.

docs/source/crud/piccolo_crud.rst
```diff
@@ -244,6 +244,66 @@ For example, to return results 11 to 20:
...
+    the usage is the same.
+
+A query which update movies with ``id`` pass in query parameter:
+
+.. code-block::
+
+    PATCH https://demo1.piccolo-orm.com/api/tables/movie/?rows_ids=1,2,3
+
+If you pass a wrong or non-existent value to the query parameters ``rows_ids``,
+no record will be changed and api response will be empty list.
```
dantownsend, line 300: For bulk update, we're using a parameter called `row_ids`, and for bulk delete we're using a parameter called `__ids`. I think it would be better if they're the same (i.e. both are `__ids`).

sinisaos: Can we stick to `rows_ids` (for both bulk delete and update) as I'm having trouble testing the FastAPI patch bulk in [here](https://github.com/sinisaos/piccolo_api/blob/6e072d29da05da5eb0284d98cf8c0389198bda89/piccolo_api/fastapi/endpoints.py#L278) with the `__ids` query parameter and I getting error. With `rows_ids` the test passes without problems.

dantownsend: To make `ids` work, you might need to modify this so it's DELETE or PATCH:

https://github.com/sinisaos/piccolo_api/blob/6e072d29da05da5eb0284d98cf8c0389198bda89/piccolo_api/fastapi/endpoints.py#L483-L499

PR #149, opened 2022-06-08
user change password endpoint

sinisaos:
Related to #148


piccolo_api/change_password/endpoints.py
```diff
@@ -0,0 +1,181 @@
...
+            else:
+                raise HTTPException(
+                    status_code=401, detail="Passwords do not match."
+                )
+
+        session_user = (
+            await SessionsBase.select(SessionsBase.user_id)
+            .where(SessionsBase.token == request.cookies.get("id"))
+            .first()
+        )
```
dantownsend, line 130: We probably don't need this - can get the Piccolo user via `request.user.user`.

piccolo_api/change_password/endpoints.py
```diff
@@ -0,0 +1,181 @@
...
+            )
+        )
+
+    async def get(
+        self, request: Request
+    ) -> t.Union[HTMLResponse, PlainTextResponse]:
+        session_user = (
+            await SessionsBase.select(SessionsBase.user_id)
+            .where(SessionsBase.token == request.cookies.get("id"))
+            .first()
```
dantownsend, line 73: We can use `request.user.user` here instead.

piccolo_api/change_password/endpoints.py
```diff
@@ -0,0 +1,181 @@
...
+            try:
+                body = await request.json()
+            except JSONDecodeError:
+                body = await request.form()
+
+        old_password = body.get("old_password", None)
+        new_password = body.get("new_password", None)
+        confirm_password = body.get("confirm_password", None)
+
+        if (not old_password) or (not new_password) or (not confirm_password):
```
dantownsend, line 95: We also need to check that the old password is correct.

For example:

```
piccolo_user = request.user.user
if not await piccolo_user.__class__.login(username=piccolo_user.username, password=old_password):
    raise HTTPException(status_code=401, detail="Incorrect password")
```

piccolo_api/change_password/endpoints.py
```diff
@@ -0,0 +1,181 @@
...
+        # to login endpoint to login again with new password
+        response = RedirectResponse(
+            url="/login/", status_code=HTTP_303_SEE_OTHER
+        )
+        response.delete_cookie("id")
+        return response
+
+
+def change_password(
+    auth_table: t.Optional[t.Type[BaseUser]] = None,
```
dantownsend, line 146: We can probably get rid of this - I don't think we use it.

piccolo_api/change_password/endpoints.py
```diff
@@ -0,0 +1,181 @@
...
+
+        old_password = body.get("old_password", None)
+        new_password = body.get("new_password", None)
+        confirm_password = body.get("confirm_password", None)
+
+        if (not old_password) or (not new_password) or (not confirm_password):
+            raise HTTPException(
+                status_code=401,
+                detail="Form is invalid. Missing one or more fields.",
+            )
```
dantownsend, line 99: If the request was via a HTML, we should return the template showing the error.

For example:

https://github.com/piccolo-orm/piccolo_api/blob/2806109eaae6093b4bdd23bd715ae9458222e268/piccolo_api/register/endpoints.py#L123-L131

dantownsend:
This looks great, thanks. Just a few minor comments.

piccolo_api/change_password/endpoints.py
sinisaos: You're right. I will add that.

piccolo_api/change_password/endpoints.py
sinisaos: You're right. I will remove that. Thanks

piccolo_api/change_password/endpoints.py
sinisaos: I can add that, but we don't use that in [register](https://github.com/piccolo-orm/piccolo_api/blob/2806109eaae6093b4bdd23bd715ae9458222e268/piccolo_api/register/endpoints.py#L111) .  I can add that to both files.

piccolo_api/change_password/endpoints.py
sinisaos: I used this to avoid use `AuthenticationMiddleware` and `SessionsAuthBackend` (as I understand correctly we need this middlewares to use `request.user.user`) because I want to protect only `change_password` endpoint not whole app. I need to see if I can change that.

dantownsend: I think it will have to be used with authentication middleware.

piccolo_api/change_password/endpoints.py
sinisaos: OK.  I will add it to both the `register` and `change_password`.

piccolo_api/change_password/endpoints.py
sinisaos: Ok. I will try to implement authentication middleware.

dantownsend:
Thanks for those changes - it looks good. I'll test it properly tomorrow.

dantownsend:
OK, I think it's almost there.

Some things I've realised. The minimum password length is hardcoded to 6. But the `_min_password_length` value in `BaseUser` might be higher than this. So ideally we will use this value instead.

We're using `BaseUser.update_password` to change the password. This currently doesn't do much password validation, so I need to open a PR for that in Piccolo.

Also, rather than redirecting straight away after changing the password, I think it's best to show the template again saying 'changed password successfully', with a 'next' button which redirects them. Otherwise it might not be clear to them that the password was changed successfully.

piccolo_api/change_password/endpoints.py
```diff
@@ -135,19 +139,23 @@ async def post(self, request: Request) -> Response:
         # after password changes we invalidate session and redirect user
         # to login endpoint to login again with new password
         response = RedirectResponse(
-            url="/login/", status_code=HTTP_303_SEE_OTHER
+            url=self._redirect_to, status_code=HTTP_303_SEE_OTHER
         )
```
sinisaos, line 143: I think this should be removed because we don't want user be able to change the redirect (by default it should be `redirect_to ="/login"` like it was before) because we are invalidating the session cookie and the user **MUST** log in again. I intentionally did not add `_redirect_to` property.

dantownsend: I see your point. The only issue is if someone has their login at a different URL. Instead of `redirect_to`, it could be `login_url` or something?

sinisaos: Well noticed! You are right. I didn't think anyone could set a different login url than` /login`. `login_url` will be great and understandable.

dantownsend:
Thanks for making those changes.

I had an idea - how about we merge the two templates (the change password and success templates). That way it's easier for the use to override them if they want using the `template_path` variable (at the moment they won't be able to override the success template unless we add another variable like `success_template_path`). A variable called 'success' can be passed in to the template, and then it will look like:

```jinja2
{% if success %}
    <p>You successfully changed your password ...</p>
{% else %}
    <!-- The login form -->
    <form>...</form>
{% endif %}
```

What do you think? I'll make the change if you like.

sinisaos:
@dantownsend Of course. Change this to what you think is best.

I have one question. Once you merge this to the master, I want to do a PR to change password in Piccolo Admin. I have a working code that works without problems with Postgres (I `npm run build` adminUI and add to the existing project and everything works fine), but it doesn't work with [example.py](https://github.com/piccolo-orm/piccolo_admin/blob/master/piccolo_admin/example.py) which uses Sqlite, even if I set `persist` to True. Do you know maybe where the problem is or you can see it later when I do PR?

dantownsend:
@sinisaos I'm not sure. Is `example.py` raising any exceptions?

sinisaos:
@dantownsend Unfortunately, there are no exceptions or anything that helps find the problem.

dantownsend:
@sinisaos Strange. I'm not sure what the problem is - maybe the proxying isn't working properly? I'll have to try it locally.

sinisaos:
@dantownsend OK. Once you merge this PR and release a new Piccolo API version I be make a PR and you can try it locally.

dantownsend:
Thanks for making those changes. I'll try it out tomorrow.

sinisaos:
Great. Thanks.

> @sinisaos I'm not sure. Is `example.py` raising any exceptions?

Actually there is error `ModuleNotFoundError: No module named 'piccolo conf'`. Do we need to change Piccolo Admin folder structure and make more like Piccolo project or app? I tried to make `piccolo_conf.py` in root adding `SQLiteEngine` in it but error stays the same. I don't know where is problem.

sinisaos:
@dantownsend I couldn't get it to work with `python -m piccolo_admin.example` or `admin_demo`. `example.py` has not changed. The only change is that I added a password change endpoint to `endpoints.py` like this. 
```python
auth_app.mount(
    path="/change-password/",
    app=AuthenticationMiddleware(
        change_password(),
        SessionsAuthBackend(), # also tried SessionsAuthBackend(session_table=session_table),
    ),
)
```
I'm probably doing something wrong, but it doesn't matter, the most important thing is that it works for you. Once you merge this PR and release a new version of the Piccolo API, I will be doing PR in Piccolo Admin and you can check it all out. Postgres work as expected.

dantownsend:
@sinisaos It's especially weird that it works OK for Postgres, but not SQLite. But yeah, don't worry about it for now - I'll test our the PR.

sinisaos:
@dantownsend Works without problems with Postgres (I `npm run build` adminUI and add to the existing project, add change password endpoint to Piccolo Admin `endpoints.py` and everything works fine) but that project has `piccolo_conf.py` which is generated with `piccolo asgi new`.

dantownsend:
@sinisaos Sorry, I haven't finished this yet. I want to update the docs slightly to mention `CSRFMiddleware`. Will do it tomorrow.

sinisaos:
@dantownsend I finally figured out. We need to use `BaseUser` and `SessionsBase` directly in `example.py` and `sandbox.py`. We just need to remove [those code lines](https://github.com/piccolo-orm/piccolo_admin/blob/master/piccolo_admin/example.py#L47-L52) from `example.py` and update `sandbox.py` to use `BaseUser` and `SessionsBase` and everything works great (I think the inherited tables required `piccolo_conf.py`, but it doesn't matter because now everything works)

Once you merge this PR I will add PR to Piccolo Admin. I also think to add possibility to create new user if you agree. Result is:


https://user-images.githubusercontent.com/30960668/173819968-85b5da5d-eaab-48e3-9096-084d4f2604f3.mp4

dantownsend:
That looks great!

sinisaos:
Thanks. Video is a bit slow, but you can see what I mean.

dantownsend:
I made a couple of changes.

1. I'm now passing in the minimum password length into the template - before it always said 6, but the actual value could be different based on `BaseUser._min_password_length`.
2. I've added extra arguments to the `change_password` endpoint - so the user can set the name of the session cookie, and tell us which sessions table to use. When the user successfully changes their password, the sessions are also deleted from the database, and not just the cookie, to provide extra safety.

A few more tests will be needed to cover these changes.

piccolo_api/change_password/endpoints.py
```diff
@@ -88,77 +103,96 @@ async def post(self, request: Request) -> Response:
...
+                response.delete_cookie(self._session_cookie_name)
+
+            session_table = self._session_table
+            if session_table:
+                # TODO - not sure if deleting or marking inactive is most
+                # appropriate. Requires research.
+                await session_table.delete().where(
+                    session_table.user_id == piccolo_user.id
+                )
+
```
sinisaos, line 188: I think deleting invalidated sessions is a great and safe idea. Why keep invalidated sessions in db when we force user to log in again and create new session? I think this should also be applied to the logout endpoint.

dantownsend: Yes, I was looking at the logout endpoint works, and that just deletes the user's session for their current device.

I was trying to find some concrete recommendations on OWASP for best practices for the following:

 * Marking sessions expired in the database vs deleting them (there's no point keeping old sessions around forever in the database, but it does keep an audit trail).
 * Whether to just delete the session for the current device when logging out / changing password, or for all of that user's devices. My feeling is logging out should be the current device only, but changing password should be all devices.

piccolo_api/change_password/endpoints.py
```diff
@@ -88,77 +103,96 @@ async def post(self, request: Request) -> Response:
...
+            return response
```
sinisaos, line 190: This should be like this 
```python
# After the password changes, we invalidate the session and
# redirect the user to the login endpoint.
response = RedirectResponse(
    url=self._login_url, status_code=HTTP_303_SEE_OTHER
)

if self._session_cookie_name:
    response.delete_cookie(self._session_cookie_name)

session_table = self._session_table
if session_table:
    # TODO - not sure if deleting or marking inactive is most
    # appropriate. Requires research.
    await session_table.delete().where(
        session_table.user_id == piccolo_user.id
    )

if body.get("format") == "html":
    return self.render_template(
        request,
        success=True,
        min_password_length=min_password_length,
        login_url=self._login_url,
    )

return response
```
because sessions are not deleted with the previous code. I just tried that.

piccolo_api/change_password/endpoints.py
```diff
@@ -192,5 +232,7 @@ class _ChangePasswordEndpoint(ChangePasswordEndpoint):
         _login_url = login_url
         _change_password_template = change_password_template
         _styles = styles or Styles()
+        _session_table = session_table
```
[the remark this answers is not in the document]
dantownsend: Thanks, I'll fix it.

piccolo_api/change_password/endpoints.py
dantownsend: I realised I messed up here. We can't show the success page after changing the password is successful - because they will already be logged out, so won't be able to view it. I think we just have to say in the change password page 'If successful, you will be redirected to the login page'.

sinisaos:
@dantownsend I agree with both ways (redirection or this new way of displaying the message directly in the template). You should merge this and release new version.

piccolo_api/change_password/endpoints.py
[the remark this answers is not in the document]
sinisaos: Reminder. You forgot to add this line.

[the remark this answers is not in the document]
dantownsend: It seems to work for me - do you still get an error?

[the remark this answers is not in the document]
dantownsend: In Piccolo admin, can we just pass in `SessionsBase`?

```python
change_password(session_table = SessionsBase)
```

We could add `SessionsBase` as a default, but my only concern is it ties this endpoint to session auth. I was hoping it could be used with other forms of auth too, but maybe it's unrealistic.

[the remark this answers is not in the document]
sinisaos: You are right. Now I tried `change_password(session_table=SessionsBase)` in Piccolo Admin and everything work fine. Sorry.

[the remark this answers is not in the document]
dantownsend: It's fine. I thought about it, and I think you're right - let's just make `SessionsBase` the default. If someone doesn't want to use it, they just set it to `None`. I think most people will use it with `SessionsBase` any way, so it's a sensible default.

What I'd like to add in a future PR is hooks, so people can perform their own logic when the password change is successful. That way people might be able to use it with other forms of auth.

dantownsend:
OK, going to release this now. Thanks for adding this.

I created a couple of follow up issues - one for adding hooks, and another for general documentation improvements (there's a lot of functionality in Piccolo API now, so need to make it as easy to get started as possible).

PR #153, opened 2022-06-18
old password -> current password

dantownsend:
Having researched some other change password forms, 'current password' makes more sense than 'old password'.


dantownsend:
@sinisaos So you're aware.

PR #158, opened 2022-06-25
enable user creation and editing

sinisaos:
Enabled user creation and editing in [Piccolo Admin](https://github.com/piccolo-orm/piccolo_admin/pull/168)


piccolo_api/crud/endpoints.py
```diff
@@ -806,18 +807,32 @@ async def post_single(
...
-                    hooks=self._hook_map, hook_type=HookType.pre_save, row=row
+        if issubclass(self.table, BaseUser):
+            self.table._validate_password(password=cleaned_data["password"])
+            cleaned_data["password"] = self.table.hash_password(
+                cleaned_data["password"]
+            )
+            model = self.pydantic_model(**cleaned_data)
+            user_row = self.table(**model.dict())
+            await user_row.save().run()
+            return Response(status_code=200)
```
dantownsend, line 818: I see your idea. You might be able to do this instead:

```python
user = await self.table.create_user(**model.dict())
json = dump_json({'id': user.id})
return CustomJSONResponse(json, status_code=201)
```

It still validates the password and hashes it. Also, it returns the data in the same format as other tables.

piccolo_api/crud/endpoints.py
```diff
@@ -1038,37 +1053,69 @@ async def patch_single(
...
-            new_row = (
-                await cls.select(exclude_secrets=self.exclude_secrets)
-                .where(cls._meta.primary_key == row_id)
-                .first()
-                .run()
-            )
-            return CustomJSONResponse(self.pydantic_model(**new_row).json())
-        except ValueError:
-            return Response("Unable to save the resource.", status_code=500)
+            return Response(status_code=200)
```
dantownsend, line 1080: I see what you're doing. I need to think about this a bit.

Maybe this simpler logic works:

```python
values = model.dict()
if model.password:
    cls._validate_password(model.password)
    values['password'] = cls.hash_password(model.password)
else:
    values.pop('password')

await cls.update(**values).where(
    cls.email == cleaned_data["email"]
).run()
return Response(status_code=200)
```

piccolo_api/crud/endpoints.py
sinisaos: This works perfectly and is much simpler. Thanks.

piccolo_api/crud/endpoints.py
sinisaos: Unfortunately I failed to implement this. I know this is a little complicated. I’ll push another commit with exceptions and if you come up with something easier to follow, it will be great.

sinisaos: I managed to change the code based on your suggestion in the last commit. I think it's much better now.

sinisaos:
@dantownsend I think we don't need [this](https://github.com/piccolo-orm/piccolo_api/blob/master/piccolo_api/crud/endpoints.py#L248-L252) and [this](https://github.com/piccolo-orm/piccolo_api/blob/master/piccolo_api/crud/endpoints.py#L346-L350) anymore.

dantownsend:
@sinisaos Yes, they can be removed - don't think they'll be used.

dantownsend:
@sinisaos I'm a bit confused why MyPy is failing in GitHub Actions. It works OK for me locally. Is it the same for you?

dantownsend:
The MyPy errors are when the latest version of fastapi is installed.

dantownsend:
@sinisaos Sorry for the delay on this - I thought we already merged it in.

sinisaos:
@dantownsend No worries. When you have time, look at the other PRs in Piccolo Api, and also in Piccolo Admin, where we have several stale PRs.

PR #160, opened 2022-06-26
initial commit for audit logs

sinisaos:
Related to #159

### DIFF
--- a/docs/source/audit_logs/tables.rst
+++ b/docs/source/audit_logs/tables.rst
@@ -1,0 +1,63 @@
+Tables
+======
+
+``audit_logs`` is a ``Piccolo`` app that records changes made by users to database tables.
+We store the audit logs in :class:`AuditLog <piccolo_api.audit_logs.tables.AuditLog>`.
+
+-------------------------------------------------------------------------------
+
+.. _AuditLogMigrations:
+
+Migrations
+----------
+
+We recommend creating ``audit_logs`` tables using migrations.
+
+You can add ``piccolo_api.audit_logs.piccolo_app`` to the ``apps`` arguments
+of the :class:`AppRegistry <piccolo.conf.apps.AppRegistry>` in ``piccolo_conf.py``.
+
+.. code-block:: bash
+
+    APP_REGISTRY = AppRegistry(
+        apps=[
+            ...
+            "piccolo_api.audit_logs.piccolo_app",
+            ...
+        ]
+    )
+
+To learn more about Piccolo apps, see the `Piccolo docs <https://piccolo-orm.readthedocs.io/en/latest/piccolo/projects_and_apps/index.html>`_.
+
+To run the migrations and create the table, run:
+
+.. code-block:: bash
+
+    piccolo migrations forwards audit_logs
+
+-------------------------------------------------------------------------------
+
+Creating them manually
+----------------------
+
+If you prefer not to use migrations, and want to create them manually, you can
+do this instead:
+
+.. code-block:: python
+
+    from piccolo_api.audit_logs.tables import AuditLog
+    from piccolo.tables import create_tables
+
+    create_tables(AuditLog, if_not_exists=True)
+
+-------------------------------------------------------------------------------
+
+Source
+------
+
+AuditLog
+~~~~~~~~~~~~
+
+.. currentmodule:: piccolo_api.audit_logs.tables
+
+.. autoclass:: AuditLog
+    :members:
--- a/piccolo_api/audit_logs/tables.py
+++ b/piccolo_api/audit_logs/tables.py
@@ -1,0 +1,111 @@
+import typing as t
+import uuid
+from datetime import datetime
+from enum import Enum
+
+from piccolo.apps.user.tables import BaseUser
+from piccolo.columns import Text, Timestamp, Varchar
+from piccolo.table import Table
+
+
+class AuditLog(Table):
+    class ActionType(str, Enum):
+        """An enumeration of AuditLog table actions type."""
+
+        CREATING = "CREATING"
+        UPDATING = "UPDATING"
+        DELETING = "DELETING"
+
+    action_time = Timestamp()
+    action_type = Varchar(choices=ActionType)
+    action_user = Varchar()
+    change_message = Text()
+
+    @classmethod
+    async def post_save_action(cls, table: t.Type[Table], user_id: int):
+        """
+        A method for tracking creating record actions.
+
+        :param table:
+            A table for which we monitor activities.
+        :param user_id:
+            The ``primary key`` of authenticated user.
+        """
+        result = cls(
+            action_time=datetime.now(),
+            action_type=cls.ActionType.CREATING,
+            action_user=cls.get_user_username(user_id),
+            change_message=f"User {cls.get_user_username(user_id)} "
+            f"create new row in {table._meta.tablename.title()} table",
+        )
+        await result.save().run()
+
+    @classmethod
+    async def post_patch_action(
+        cls,
+        table: t.Type[Table],
+        row_id: t.Union[str, uuid.UUID, int],
+        user_id: int,
+    ):
+        """
+        A method for tracking updating record actions.
+
+        :param table:
+            A table for which we monitor activities.
+        :param row_id:
+            The ``primary key`` of the table for which we
+            monitor activities.
+        :param user_id:
+            The ``primary key`` of authenticated user.
+        """
+        result = cls(
+            action_time=datetime.now(),
+            action_type=cls.ActionType.UPDATING,
+            action_user=cls.get_user_username(user_id),
+            change_message=f"User {cls.get_user_username(user_id)} update row "
+            f"{row_id} in {table._meta.tablename.title()} table",
+        )
+        await result.save().run()
+
+    @classmethod
+    async def post_delete_action(
+        cls,
+        table: t.Type[Table],
+        row_id: t.Union[str, uuid.UUID, int],
+        user_id: int,
+    ):
+        """
+        A method for tracking deletion record actions.
+
+        :param table:
+            A table for which we monitor activities.
+        :param row_id:
+            The ``primary key`` of the table for which we
+            monitor activities.
+        :param user_id:
+            The ``primary key`` of authenticated user.
+        """
+        result = cls(
+            action_time=datetime.now(),
+            action_type=cls.ActionType.DELETING,
+            action_user=cls.get_user_username(user_id),
+            change_message=f"User {cls.get_user_username(user_id)} delete row "
+            f"{row_id} in {table._meta.tablename.title()} table",
+        )
+        await result.save().run()
+
+    @classmethod
+    def get_user_username(cls, user_id: int) -> str:
+        """
+        Returns the username of authenticated user.
+
+        :param user_id:
+            The ``primary key`` of authenticated user.
+        """
+        user = (
+            BaseUser.select(BaseUser.username)
+            .where(BaseUser._meta.primary_key == user_id)
+            .first()
+            .run_sync()
+        )
+        return user["username"]
--- a/piccolo_api/crud/endpoints.py
+++ b/piccolo_api/crud/endpoints.py
@@ -27,6 +27,7 @@ from starlette.requests import Request
 from starlette.responses import JSONResponse, Response
 from starlette.routing import Route, Router
 
+from piccolo_api.audit_logs.tables import AuditLog
 from piccolo_api.crud.hooks import (
     Hook,
     HookType,
@@ -812,6 +813,12 @@ class PiccoloCRUD(Router):
                 row = await execute_post_hooks(
                     hooks=self._hook_map, hook_type=HookType.pre_save, row=row
                 )
+            try:
+                await AuditLog.post_save_action(
+                    self.table, user_id=request.user.user_id
+                )
+            except AssertionError:
+                pass
             response = await row.save().run()
             json = dump_json(response)
             # Returns the id of the inserted row.
@@ -1060,6 +1067,12 @@ class PiccoloCRUD(Router):
             await cls.update(values).where(
                 cls._meta.primary_key == row_id
             ).run()
+            try:
+                await AuditLog.post_patch_action(
+                    cls, row_id=row_id, user_id=request.user.user_id
+                )
+            except AssertionError:
+                pass
             new_row = (
                 await cls.select(exclude_secrets=self.exclude_secrets)
                 .where(cls._meta.primary_key == row_id)
@@ -1089,6 +1102,12 @@ class PiccoloCRUD(Router):
             await self.table.delete().where(
                 self.table._meta.primary_key == row_id
             ).run()
+            try:
+                await AuditLog.post_delete_action(
+                    self.table, row_id=row_id, user_id=request.user.user_id
+                )
+            except AssertionError:
+                pass
             return Response(status_code=204)
         except ValueError:
             return Response("Unable to delete the resource.", status_code=500)
--- a/docs/source/audit_logs/commands.rst
+++ b/docs/source/audit_logs/commands.rst
@@ -1,0 +1,16 @@
+Commands
+========
+
+If you've registered the ``audit_logs`` app in your ``piccolo_conf.py`` file
+(see the :ref:`migrations docs <AuditLogMigrations>`), it gives you access to a
+custom command.
+
+clean
+-----
+
+If you run the following on the command line, it will delete all logs
+from the database.
+
+.. code-block:: bash
+
+    piccolo audit_logs clean
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -41,6 +41,13 @@ ASGI app, covering authentication, security, and more.
    ./advanced_auth/index
 
 .. toctree::
+   :caption: Audit logs
+   :maxdepth: 1
+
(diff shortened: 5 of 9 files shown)


docs/source/audit_logs/tables.rst
```diff
@@ -0,0 +1,63 @@
...
+We store the audit logs in :class:`AuditLog <piccolo_api.audit_logs.tables.AuditLog>`.
+
+-------------------------------------------------------------------------------
+
+.. _AuditLogMigrations:
+
+Migrations
+----------
+
+We recommend creating ``audit_logs`` tables using migrations.
```
dantownsend, line 14: creating ``audit_logs``-> creating the ``audit_logs``

docs/source/audit_logs/tables.rst
```diff
@@ -0,0 +1,63 @@
...
+
+If you prefer not to use migrations, and want to create them manually, you can
+do this instead:
+
+.. code-block:: python
+
+    from piccolo_api.audit_logs.tables import AuditLog
+    from piccolo.tables import create_tables
+
+    create_tables(AuditLog, if_not_exists=True)
```
dantownsend, line 50: I added `create_db_tables_sync` in a recent version of Piccolo - might be better using this. `create_tables` was a bit of an anomaly because it's sync first, with no async equivalent, which is why I added `create_db_tables` and `create_db_tables_sync` to replace it.

docs/source/audit_logs/tables.rst
```diff
@@ -0,0 +1,63 @@
...
+
+-------------------------------------------------------------------------------
+
+Source
+------
+
+AuditLog
+~~~~~~~~~~~~
```
dantownsend, line 58: Reduce the underline length to match the text - not a big issue, but some Restructured Text linters complain.

piccolo_api/crud/endpoints.py
```diff
@@ -812,6 +813,12 @@ async def post_single(
                 row = await execute_post_hooks(
                     hooks=self._hook_map, hook_type=HookType.pre_save, row=row
                 )
+            try:
+                await AuditLog.post_save_action(
+                    self.table, user_id=request.user.user_id
+                )
+            except AssertionError:
```
dantownsend, line 820: Where are the `AssertionErrors` being raised?

piccolo_api/audit_logs/tables.py
```diff
@@ -0,0 +1,111 @@
...
+        """An enumeration of AuditLog table actions type."""
+
+        CREATING = "CREATING"
+        UPDATING = "UPDATING"
+        DELETING = "DELETING"
+
+    action_time = Timestamp()
+    action_type = Varchar(choices=ActionType)
+    action_user = Varchar()
+    change_message = Text()
```
dantownsend, line 22: I think one really nice thing would be to be able to which changes were made. There could be a JSON field which would have something like `{'name': 'bob'}` if the name field was changed to 'bob'.

piccolo_api/crud/endpoints.py
```diff
@@ -1060,6 +1067,12 @@ async def patch_single(
             await cls.update(values).where(
                 cls._meta.primary_key == row_id
             ).run()
+            try:
+                await AuditLog.post_patch_action(
+                    cls, row_id=row_id, user_id=request.user.user_id
+                )
```
dantownsend, line 1073: I think it should all be optional. So PiccoloCRUD should have a new argument which is `audit_log: t.Optional[t.Type[AuditLog]] = None`. And then this line would be:

```python
if self.audit_table:
    await AuditLog.post_patch_action(
        cls, row_id=row_id, user_id=request.user.user_id
     )
```

Otherwise when we release this it will be a breaking change - anyone without the `AuditLog` table installed will have their code break. Also, it allows people to disable this functionality if they don't want to track changes.

piccolo_api/audit_logs/tables.py
```diff
@@ -0,0 +1,111 @@
...
+            A table for which we monitor activities.
+        :param user_id:
+            The ``primary key`` of authenticated user.
+        """
+        result = cls(
+            action_time=datetime.now(),
+            action_type=cls.ActionType.CREATING,
+            action_user=cls.get_user_username(user_id),
+            change_message=f"User {cls.get_user_username(user_id)} "
+            f"create new row in {table._meta.tablename.title()} table",
```
dantownsend, line 39: It would be better if this said which row was created.

piccolo_api/audit_logs/tables.py
```diff
@@ -0,0 +1,111 @@
...
+        """
+        A method for tracking creating record actions.
+
+        :param table:
+            A table for which we monitor activities.
+        :param user_id:
+            The ``primary key`` of authenticated user.
+        """
+        result = cls(
+            action_time=datetime.now(),
```
dantownsend, line 35: This can be omitted - it should default to `now`.

piccolo_api/audit_logs/tables.py
```diff
@@ -0,0 +1,111 @@
...
+            action_type=cls.ActionType.CREATING,
+            action_user=cls.get_user_username(user_id),
+            change_message=f"User {cls.get_user_username(user_id)} "
+            f"create new row in {table._meta.tablename.title()} table",
+        )
+        await result.save().run()
+
+    @classmethod
+    async def post_patch_action(
```
dantownsend, line 44: I'd consider naming these methods slightly differently. I know what you mean by them, and they are correct. But having `post_patch` could be confusing - someone might misread it as being about posts and not patches.

Instead of `post_`, how about just `record_`? So it would be `record_patch_action`?

We will never have a `pre_` version, so no need to distinguish with `post_`.

dantownsend:
@sinisaos I think this is really cool - thanks. Left some comments.

piccolo_api/audit_logs/tables.py
sinisaos: Great idea. I will try to implement it.

piccolo_api/crud/endpoints.py
sinisaos: You're right. I will change that.

piccolo_api/crud/endpoints.py
sinisaos: I used this to prevent `AssertionError` from `AuthenticationMiddleware` because we should be able to use `PiccoloCRUD` without session authentication which `AuditLog` depends on (if we use e.g. OAuth in FastAPI docs without this, we get `AssertionError: AuthenticationMiddleware must be installed to access request.user` when trying to save records)

piccolo_api/audit_logs/tables.py
sinisaos: You're right. I will change that.

piccolo_api/crud/endpoints.py
dantownsend: I see what you mean. I think the only problem is it ties it really closely to FastAPI's internals, which could change in the future. How about we catch all exceptions and log it? Something like:

```python
try:
     await AuditLog....
except Exception as exception:
    logger.log(exception)
```

piccolo_api/crud/endpoints.py
```diff
@@ -218,6 +219,7 @@ def __init__(
             }
         else:
             self._hook_map = None  # type: ignore
+        self.audit_log_table = audit_log_table or AuditLog
```
dantownsend, line 222: There's no need to the `or` here - the user should be allowed to set it to `None`.

sinisaos: But without that we cannot add records to the `AuditLog` table (or I doing something wrong). If the user does not want to use the `audit_logs` app, just does not register it in `APP_REGISTRY` and can use `PiccoloCRUD` without `audit_logs`.

dantownsend: Even if people don't register the `audit_logs` app, `PiccoloCRUD` will still be trying to make database queries to the `AuditLog` table, so we need to be able to set it to `None` to prevent this. I think when people use `PiccoloCRUD` standalone within their apps they probably won't need the audit logs - it's mostly for Piccolo Admin.

In `create_admin` we could add an `enable_auditing` argument - any tables listed will have auditing switched on in `PiccoloCRUD`.

```python
def create_admin(..., audit_log_table: AuditLog = AuditLog, enable_auditing: t.List[t.Type[Table]] = [])
```

sinisaos: Good point. Later today I will try to do that in Piccolo Admin.

sinisaos: You are right about this. We need to add a few lines of code to Piccolo Admin and everything works great and we can choose which table we want audit logs for in future `enable_auditing` argument in `create_admin`.

piccolo_api/audit_logs/tables.py
```diff
@@ -0,0 +1,118 @@
...
+        :param row_id:
+            The ``primary key`` of the table for which we
+            monitor activities.
+        :param user_id:
+            The ``primary key`` of authenticated user.
+        """
+        result = cls(
+            action_type=cls.ActionType.DELETING,
+            action_user=cls.get_user_username(user_id),
+            change_message=f"User {cls.get_user_username(user_id)} delete row "
```
dantownsend, line 99: I'd be tempted to get rid of the change message all together.

One option is just storing everything in JSON.

For delete, the json can be:

```
// Maps to a list, so we can also use it for a bulk delete endpoint.
{
    "deleted_ids": [1, 2]
}
```

Likewise, for the save, it could be:

```
// Maps to a list, so we can also use it for a bulk creation endpoint.
{
    "created_ids": [1, 2]
}
```

Or we have a field called `effected_rows` which is an integer or array of integers.

It's just about making it easy to query - imagine you wanted to know who deleted row 42. You would have to something like:

```python
await AuditLog.select().where(
    AuditLog.action_type == AuditLog.ActionType.DELETING,
    AuditLog.change_message.like('%42%')
)
```

It's quite imprecise, ad would be better to just do:

```python
await AuditLog.select(AuditLog.action_user).where(
    AuditLog.action_type == AuditLog.ActionType.DELETING,
    AuditLog.effected_rows.any(42)
)
```

piccolo_api/audit_logs/tables.py
```diff
@@ -0,0 +1,118 @@
...
+from piccolo.table import Table
+
+
+class AuditLog(Table):
+    class ActionType(str, Enum):
+        """An enumeration of AuditLog table actions type."""
+
+        CREATING = "CREATING"
+        UPDATING = "UPDATING"
+        DELETING = "DELETING"
```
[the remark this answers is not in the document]
sinisaos: Ok. I will change that.

piccolo_api/audit_logs/tables.py
sinisaos: > I'd be tempted to get rid of the change message all together.
> 
> One option is just storing everything in JSON.
> 
> For delete, the json can be:
> 
> ```
> // Maps to a list, so we can also use it for a bulk delete endpoint.
> {
>     "deleted_ids": [1, 2]
> }
> ```
> 
> Likewise, for the save, it could be:
> 
> ```
> // Maps to a list, so we can also use it for a bulk creation endpoint.
> {
>     "created_ids": [1, 2]
> }
> ```

But all actions are for endpoints that work with a single record. We don't have the `bulk_delete` option yet ([PR here](https://github.com/piccolo-orm/piccolo_api/pull/145)) and we don't have `bulk_creation` at all. Audit logs would be primarily used in Piccolo Admin and there we delete multiple records by looping through the [delete_single](https://github.com/piccolo-orm/piccolo_admin/blob/2fbdec3a63c919bc4c140b75ec5469b4921e78d0/admin_ui/src/views/RowListing.vue#L417) endpoint.

> Or we have a field called `effected_rows` which is an integer or array of integers.

That would be possible if we have only `primary_key` type `Serial`. But we also have `primary_key` type `string` and `UUID`, and we can't use `Integer` column for that (which will be great for precise filtering). We need to use `Varchar` and query would be very similar for `effected_row` and `change_message` using `like`.

> It's just about making it easy to query - imagine you wanted to know who deleted row 42. You would have to something like:
> 
> ```python
> await AuditLog.select().where(
>     AuditLog.action_type == AuditLog.ActionType.DELETING,
>     AuditLog.change_message.like('%42%')
> )
> ```
> 
> It's quite imprecise, ad would be better to just do:
> 
> ```python
> await AuditLog.select(AuditLog.action_user).where(
>     AuditLog.action_type == AuditLog.ActionType.DELETING,
>     AuditLog.effected_rows.any(42)
> )
> ```
Here is screenshot with new columns.

changed_columns

I don't think we should remove the `change_message` column because it nicely describes the action the user has taken. Filtering is maybe imprecise, but we can't use the Integer column for `effected_row`.

> Similarly, table name should probably be a column too, to make it easier to filter out the changes on a certain table. I'd be tempted to add `table_name` and `table_class` - again, just to make it easy to filter. What do you think?

I added `table_name`, but I don't understand what the `table_class` column would be for and what it should go into that column?

dantownsend: Good points.

> But all actions are for endpoints that work with a single record. We don't have the bulk_delete option yet (https://github.com/piccolo-orm/piccolo_api/pull/145) and we don't have bulk_creation at all. Audit logs would be primarily used in Piccolo Admin and there we delete multiple records by looping through the [delete_single](https://github.com/piccolo-orm/piccolo_admin/blob/2fbdec3a63c919bc4c140b75ec5469b4921e78d0/admin_ui/src/views/RowListing.vue#L417) endpoint.

You're right, there's no bulk action endpoints at the moment. But I was just thinking about the future. Imagine someone deletes 1000 rows - we would then have to insert 1000 rows into the audit table.

> That would be possible if we have only primary_key type Serial. But we also have primary_key type string and UUID, and we can't use Integer column for that (which will be great for precise filtering). We need to use Varchar and query would be very similar for effected_row and change_message using like.

Good point - the primary key might not be an integer. Storing them all as strings is OK.

> I don't think we should remove the change_message column because it nicely describes the action the user has taken. Filtering is maybe imprecise, but we can't use the Integer column for effected_row.

Maybe `change_message` could be achieved using `get_readable` instead.

> I added table_name, but I don't understand what the table_class column would be for and what it should go into that column?

I was just thinking that if there's a table class called `MyTable`, and it refers to the database table `my_table`, maybe we would want to store them both. But it's not that important. Just need to decide which one is best to store - `my_table` or `MyTable`.

sinisaos: > You're right, there's no bulk action endpoints at the moment. But I was just thinking about the future. Imagine someone deletes 1000 rows - we would then have to insert 1000 rows into the audit table.

I agree, but we don’t currently have the classic bulk delete in Piccolo Admin (we loop through `delete_single` endpoint) and audit logs would primarily be used in Piccolo Admin (page size limit is 100 per page, which is not so bad).
Once we have the proper `bulk_delete`, we can change the delete action for the audit logs to something like this

```python
# in piccolo_api.crud.endpoints.py
@apply_validators
async def delete_bulk(
self, request: Request, params: t.Optional[t.Dict[str, t.Any]] = None
) -> Response:
    """
    Bulk deletes rows - query parameters are used for filtering.
    """
    params = self._clean_data(params) if params else {}
    split_params = self._split_params(params)
    split_params_ids = split_params.ids.split(",")
  
    try:
        query: t.Union[
            Select, Count, Objects, Delete
        ] = self.table.delete()
        try:
            # Serial or UUID primary keys enabled in query params
            value_type = self.table._meta.primary_key.value_type
            ids = [value_type(item) for item in split_params_ids]
            query_ids = query.where(
                self.table._meta.primary_key.is_in(ids)
            )
            query = self._apply_filters(query_ids, split_params)
            # this is not tested but it should be work
            # in ids list can be one or more items 
            deleted_rows_map = {"deleted_pks": ids}
            try:
                if self.audit_log_table:
                    await self.audit_log_table.record_bulk_delete_action(
                        self.table,
                        deleted_rows_pk=deleted_rows_map,
                        user_id=request.user.user_id,
                    )
            except Exception as exception:
                logger.log(msg=f"{exception}", level=logging.WARNING)
        except ValueError:
            query = self._apply_filters(query, split_params)
    except MalformedQuery as exception:
         return Response(str(exception), status_code=400)

    await query.run()
    return Response(status_code=204)

# in piccolo_api.audit_logs.tables.py

class AuditLog(Table):
    ...
    deleted_rows = JSON()
    ...
    @classmethod
    async def record_delete_action(
        cls,
        table: t.Type[Table],
        deleted_rows_pk: t.Dict[str, t.Any],
        user_id: int,
    ):
        result = cls(
            action_type=cls.ActionType.deleting,
            action_user=cls.get_user_username(user_id),
            deleted_rows=deleted_rows_pk,
        )
        await result.save().run()

```
But for now we don't have that option because we're looping through the `delete_single` endpoint and we can't store multiple IDs in a row like this
```vue
// Maps to a list, so we can also use it for a bulk delete endpoint.
{
    "deleted_ids": [1, 2]
}
```
 We should stick to what we have at the moment, and later we can always change (when we have `bulk_delete` we will have to make changes also in Piccolo Admin).

> I was just thinking that if there's a table class called `MyTable`, and it refers to the database table `my_table`, maybe we would want to store them both. But it's not that important. Just need to decide which one is best to store - `my_table` or `MyTable`.

I think `table_name` is a better option because it's `Varchar` and we can use the filter sidebar to search and filter records.

sinisaos:
Remainder: when #145 is merged we can finish this.

PR #161, opened 2022-07-03
improve http status codes

dantownsend:
Some of the HTTP status codes were correct - they were returning 401, which means permission denied, when 422 is more appropriate, as they are validation errors.

This is important for Piccolo Admin, as we configure axios to intercept any 401 errors, and redirect to the login page.


dantownsend:
@sinisaos We need this change, because that's why in this PR it was redirecting to the login page if there was an error in the change password form:

https://github.com/piccolo-orm/piccolo_admin/pull/168

sinisaos:
@dantownsend This is great and should work.

dantownsend:
@sinisaos Cool - I'll release this, then will release the change password feature in Piccolo Admin.

Issue #163, opened 2022-07-22
Small bug


sinisaos:
I find a small bug. After the last changes in Piccolo ORM `Table.__init__` `ignore_missing` argument is now `_ignore_missing` and we have to change [this line ](https://github.com/piccolo-orm/piccolo_api/blob/af7043b064081343c9e46eaef5750896edf15c04/piccolo_api/crud/endpoints.py#L850)  to `row = self.table(_ignore_missing=True)` otherwise we get 500 Internal Server Error from `get_new` method. If you want I can change it.

dantownsend:
Yes - if you could create a PR that would be great, thanks!

I didn't realise we were using `ignore_missing` in Piccolo API. I try and avoid breaking changes, it's my mistake.

PR #164, opened 2022-07-22
fix ignore_missing argument

sinisaos:
Related to #163


dantownsend:
Thanks. Can you just update `requirements.txt` so the minimum Piccolo version is the latest one?

dantownsend:
Perfect, thanks. Will release as soon as the CI finishes.

sinisaos:
No problem. I accidentally ran into a bug while I was working on uploading images and couldn't save a new record.

PR #167, opened 2022-07-28
Request Context Inside Hook Callbacks

AnthonyArmour:
Added implementation for passing request context to hook callbacks. Wrote tests for it. Hope it is suitable.
Addresses [this](https://github.com/piccolo-orm/piccolo_api/discussions/154) discussion.

### DIFF
--- a/tests/crud/test_hooks.py
+++ b/tests/crud/test_hooks.py
@@ -63,11 +63,7 @@ async def failing_hook(row_id: int):
     raise Exception("hook failed")
 
 
-def non_coroutine_hook(row: Movie):
-    # TODO: needed for coverage?
-    pass
-
-
+# TODO - add test for a non-async hook.
 class TestPostHooks(TestCase):
     def setUp(self):
         Movie.create_table(if_not_exists=True).run_sync()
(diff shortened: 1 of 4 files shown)


dantownsend:
@AnthonyArmour Thanks a lot for this. I had a look through, and it looks good.

I'll do a proper review later today.

tests/crud/test_hooks.py
```diff
@@ -39,17 +40,66 @@ async def look_up_existing(row_id: int, values: dict):
...
+    return values
+
+
+async def additional_name_details(row: Movie, request: Request):
+    director = request.query_params.get("director_name", "")
+    row["name"] = f"{row.name} ({director})"
+    return row
+
+
+async def test_raises(row_id: int, request: Request):
```
dantownsend, line 57: I think the tests are failing due to this.

It's because we use pytest as the test runner, and it thinks that any function starting with `test_` is a unittest. If it's renamed to something like `raises_exception` instead, then it should work.

AnthonyArmour: Sorry about that dumb mistake, got it fixed.

dantownsend: No worries - thanks!

PR #170, opened 2022-08-15
Support for default ACL and Content-type according to the file added

sumitsharansatsangi:
In these commit, I have added the support for `default ACL` and `content-type` addition similar to `django-storages` For AWS S3.

Currently, its support for private access only. But, now the default is private, and user can assign a valid ACL to override the behaviour.

Currently, While file upload content-type has not been assigned, which is by default assigned by AWS S3 as binary/octet-stream.
 But now, I have specified the auto-assignment  of the content type according to the file extension.

Also added support for max-cache-age.


dantownsend:
@sumitsharansatsangi Thanks a lot for this. We should probably add `.idea` to `.gitignore`.

sumitsharansatsangi:
Added `.idea` to `.gitignore`

PR #171, opened 2022-08-16
 Support for ACL, File Content Type, Cache Max Age, and user Metadata added

sumitsharansatsangi:
Currently, The visibility and accessibility of the file was private. But now user can define the visibility and accessibility (default is private). 

Currently, No any file content-type was assigned by the piccolo-api, but now it is automatically assigned by the piccolo-api.

Currently,  There is no option for defining the file cache age, but now it can be defined.

Currently, there was no support for the user defined metadata for the file, but now it can be defined.


piccolo_api/media/s3.py
```diff
@@ -38,12 +42,22 @@ def __init__(
...
         :param bucket_name:
             Which S3 bucket the files are stored in.
-        :param folder:
+        :param folder_name:
             The files will be stored in this folder within the bucket. S3
             buckets don't really have folders, but if ``folder`` is
             ``'movie_screenshots'``, then we store the file at
             ``'movie_screenshots/my-file-abc-123.jpeg'``, to simulate it being
             in a folder.
+        :param cache_max_age
```
dantownsend, line 51: Minor nit-pick - there's a colon missing here.

piccolo_api/media/s3.py
```diff
@@ -38,12 +42,22 @@ def __init__(
...
+            It takes the value in second
+            For example::
+                cache_max_age=86400
+                This will keep the cache for 24 hour.
+        :param default_acl
```
dantownsend, line 56: And here

sumitsharansatsangi: Corrected, Please do merge and release.

sinisaos: @sumitsharansatsangi Can you run `./scripts/lint.sh` because you have 3 long lines (line E501 too long) and 1 Mypy error (incompatible types in  assignment). If you don't fix that linter will complain and linter in GH actions won't pass.

sumitsharansatsangi: Ok, let me do.

sumitsharansatsangi: Corrected all the linter error,

piccolo_api/shared/middleware/junction.py
```diff
@@ -15,12 +17,14 @@ def __init__(self, *routers: Router) -> None:
     async def __call__(self, scope: Scope, receive: Receive, send: Send):
         for router in self.routers:
             try:
-                asgi = await router(scope, receive=receive, send=send)
+                response: t.Any | None = await router(
```
sinisaos: Just a tip. In future, you must be careful to use syntax that will satisfy the tests in `Py3.7`, `Py3.8`, `P3.9`, and `Py3.10`. I run your branch and everything should be fine, just remove the `import typing as t` from file because is unused.

sumitsharansatsangi: Yeah, I'll be careful. Done the changes. Hope everything will be good now. Please have a look and merge it.

sinisaos: Great. You have to wait @dantownsend for merging.

sumitsharansatsangi:
@dantownsend , Please merge this.

piccolo_api/media/s3.py
```diff
@@ -22,7 +23,10 @@ def __init__(
         column: t.Union[Text, Varchar, Array],
         bucket_name: str,
         folder_name: str,
+        cache_max_age: t.Optional[int] = None,
+        default_acl: str = "private",
```
dantownsend, line 27: Maybe we should set this to ``None`` by default?

piccolo_api/media/s3.py
```diff
@@ -127,13 +144,24 @@ def store_file_sync(
         A sync wrapper around :meth:`store_file`.
         """
         file_key = self.generate_file_key(file_name=file_name, user=user)
-
+        extension = file_key.rsplit(".", 1)[-1]
         client = self.get_client()
+        metadata: t.Dict[str, t.Any] = {
+            "ACL": self.default_acl,
+            "ContentDisposition": "inline",
```
dantownsend, line 151: Should we `ContentDisposition` configurable, by adding a `content_disposition` parameter?

dantownsend:
I think it's almost there.

I left a couple of comments.

My only other concern is before we were assuming the bucket would be private, so all of the URLs are signed. But now it could be a public bucket, so there's no need to sign the URLs. In the future we could add something like a `sign_urls` parameter.

piccolo_api/media/s3.py
dantownsend: My reasoning for this is some S3 compatible storage providers might not support ACLs. And Amazon uses private as a default anyway if it's not explicitly set.

dantownsend:
If you like I'll make the last couple of changes, and will release it.

piccolo_api/media/s3.py
sumitsharansatsangi: Yeah, we can do. No issue from my side.

sumitsharansatsangi:
Please do the change what you feel necessary and recommends and then do release.

dantownsend:
Merged in https://github.com/piccolo-orm/piccolo_api/pull/177

PR #177, opened 2022-08-18
Pr171 tweaks

dantownsend:
Based on https://github.com/piccolo-orm/piccolo_api/pull/171

I made two modifications. The main thing is there's now a single argument called `upload_metadata` for passing in the various S3 options. S3 has so many options available:

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/customizations/s3.html#boto3.s3.transfer.S3Transfer

If we just let the user pass in a dictionary instead, we cover all future use cases.

The other modification was adding a `sign_urls` argument. So if it's being used with a public bucket we're not wasting resources generating signed URLs when they're not required.


dantownsend:
@sumitsharansatsangi I made a couple of modifications.  If you agree with the changes, I'll merge it in.

sumitsharansatsangi:
I agree with the changes, please merge it.

dantownsend:
@sumitsharansatsangi OK, thanks - will release it as soon as the tests pass.

sumitsharansatsangi:
Please add one change, add support for user metadata , as I have added.

sumitsharansatsangi:
Upload_metadata will go with system metadata, not with user metadata

dantownsend:
@sumitsharansatsangi User metadata should still work if you do this:

```python
S3MediaStorage(
    ...,
    upload_metadata={'Metadata': {'myfield': 'abc123'}}
)
```

sumitsharansatsangi:
If you are sure, then please go ahead. I am not aware of this.

sumitsharansatsangi:
You were true. Please merge it, all checks passed.

Issue #186, opened 2022-09-01
`PiccoloCRUD` should handle unique exceptions better


dantownsend:
Related to https://github.com/piccolo-orm/piccolo_admin/issues/167

PR #187, opened 2022-09-01
add `db_exception_handler`

dantownsend:
Fixes https://github.com/piccolo-orm/piccolo_api/issues/186

- [x] Write tests


sinisaos:
@dantownsend I am not able to try this, but it looks great.

dantownsend:
@sinisaos Thanks. I'll get it merged.

I was going to wait until we have generic database exceptions in Piccolo, but it's quite a bit of work, so I think this is a good solution for now.

PR #200, opened 2022-11-05
Support for custom callbacks

haffi96:


dantownsend:
If it's possible without too much hassle, it would be cool if we could move the actions code down to `PiccoloCRUD`, and expose it via `FastAPIWrapper`. It just means someone could potentially use this functionality in a standalone API.

haffi96:
> If it's possible without too much hassle, it would be cool if we could move the actions code down to `PiccoloCRUD`, and expose it via `FastAPIWrapper`. It just means someone could potentially use this functionality in a standalone API.

Sure. That sounds like a good idea

Issue #202, opened 2022-11-15
Support async validator functions


dantownsend:
Currently, `PiccoloCRUD` validators must be standard Python functions. I think it's useful if they can be async too.

Now that Piccolo Admin lets you register validators too, it's more likely that async validators will be needed by users.

PR #203, opened 2022-11-15
support async validators in `PiccoloCRUD`

dantownsend:
Resolves https://github.com/piccolo-orm/piccolo_api/issues/202


dantownsend:
Tests are failing because the latest Starlette version replaced `requests` in their test client with `httpx`, and they behave quite a bit differently with redirects and cookies.

PR #206, opened 2022-11-18
fix codeql warnings

sinisaos:
Related to #201. Try to fix the codeql warning.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -16,8 +16,6 @@ import sys
 
 sys.path.insert(0, os.path.abspath("../.."))
 
-autoclass_content = "both"
-
 # -- Project information -----------------------------------------------------
 
 project = "Piccolo API"
--- a/piccolo_api/change_password/endpoints.py
+++ b/piccolo_api/change_password/endpoints.py
@@ -235,7 +235,9 @@ def change_password(
     )
 
     directory, filename = os.path.split(template_path)
-    environment = Environment(loader=FileSystemLoader(directory))
+    environment = Environment(
+        loader=FileSystemLoader(directory), autoescape=True
+    )
     change_password_template = environment.get_template(filename)
 
     class _ChangePasswordEndpoint(ChangePasswordEndpoint):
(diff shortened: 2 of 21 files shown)


piccolo_api/media/local.py
```diff
@@ -95,7 +95,7 @@
             with open(path, "wb") as new_file:
                 shutil.copyfileobj(file, new_file)
                 if file_permissions is not None:
-                    os.chmod(path, 0o640)
+                    os.chmod(path, 0o644)
```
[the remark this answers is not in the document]
sinisaos: @dantownsend This is the only warning left so I think this might work. I think the previous permission is also ok but I change your permission to `644` based on [this](https://www.namecheap.com/support/knowledgebase/article.aspx/400/205/file-permissions/), but that is not enough either. Any idea what file permission would satisfy `codeql` so that the Piccolo API code would still work. Thanks.

dantownsend:
@sinisaos Thanks a lot for fixing these.

I need to think a bit about the permissions. I think it's OK though, as we give the user the ability to override them anyway.

sinisaos:
@dantownsend Yes exactly, and the default file permissions on a Unix system are often 600 or 644 anyway.

docs/source/conf.py
```diff
@@ -16,8 +16,6 @@
 
 sys.path.insert(0, os.path.abspath("../.."))
 
-autoclass_content = "both"
```
dantownsend, line 19: Good catch - I didn't realise this was duplicated.

piccolo_api/change_password/endpoints.py
```diff
@@ -235,7 +235,9 @@ def change_password(
     )
 
     directory, filename = os.path.split(template_path)
-    environment = Environment(loader=FileSystemLoader(directory))
+    environment = Environment(
+        loader=FileSystemLoader(directory), autoescape=True
```
dantownsend, line 239: I've tested these with `autoescape=True`, and it seems to work OK still.

piccolo_api/media/local.py
```diff
@@ -95,7 +95,7 @@ def save():
             with open(path, "wb") as new_file:
                 shutil.copyfileobj(file, new_file)
                 if file_permissions is not None:
-                    os.chmod(path, 0o640)
+                    os.chmod(path, 0o600)
```
dantownsend, line 98: I think we have a mistake here (it was pre-existing). Should be:

```python
if file_permissions is not None:
    os.chmod(path, file_permissions)
```

sinisaos: Great catch. `file_permissions` are already defined in constructor. I didn't notice it at all.

Issue #135, opened 2022-01-27
Add screen grab of Swagger docs to `FastAPIWrapper` page


dantownsend:
Having a screen grab showing the generated Swagger docs would be nice on this page:

https://piccolo-api.readthedocs.io/en/latest/fastapi/index.html

The FastAPIWrapper / PiccoloCRUD combo is one of the most powerful things in the Piccolo ecosystem, and a screen grab will make it clearer the functionality it gives you.

<img width="755" alt="Screenshot 2022-01-27 at 12 05 03" src="https://user-images.githubusercontent.com/350976/151355547-3fd03254-1863-4eb8-b6cc-55cc874cef35.png">

That page is also unusual in that it uses `Task` as an example - we should change it to use `Movie`, as we can get example code / screen grabs from Piccolo Admin.

sinisaos:
@dantownsend You can close this issue because it's done.

PR #220, opened 2023-04-02
the same table names in the docs

sinisaos:
Related to #135


dantownsend:
@sinisaos Thanks for this.

Issue #150, opened 2022-06-17
Documentation improvements


dantownsend:
Now that Piccolo API has a fairly broad set of features, it's important to improve upon the documentation.

I think the following needs adding:

- [x] A 'basics' page - explaining how ASGI middleware works
- [x] Make sure all of the examples are easy to understand
- [x] Improve JWT and Token Auth docs - they don't document the endpoints properly

sinisaos:
@dantownsend You can close this issue because it's done.

PR #221, opened 2023-04-03
improved JWT and Token auth docs with examples

sinisaos:
Related to #150


dantownsend:
@sinisaos Thanks for this.

Issue #223, opened 2023-04-06
Add `allow_unauthenticated` to `JWTMiddleware`


dantownsend:
### Discussed in https://github.com/piccolo-orm/piccolo_api/discussions/222

<div type='discussions-op-text'>

<sup>Originally posted by **Fidel-C** April  6, 2023</sup>
When using JWT, Is there a way to exempt certain endpoints from requiring token in the headers?
I'd also like to know if there's a discord channel or something to join the community.</div>

PR #224, opened 2023-04-06
added `allow_unauthenticated` option to `JWTMiddleware`

dantownsend:
Resolves https://github.com/piccolo-orm/piccolo_api/issues/223


sinisaos (approved):
This is great.

dantownsend:
@sinisaos Thanks.

The JWT middleware is different to the other ones as it's not based on Starlette's `AuthenticationBackend`. Not sure if it's worth refactoring at some point, but probably OK for now.

PR #226, opened 2023-04-13
add visible_paths to JWTMiddleware

sinisaos:
Related to [discussion](https://github.com/piccolo-orm/piccolo_api/discussions/222).

### DIFF
--- a/piccolo_api/jwt_auth/middleware.py
+++ b/piccolo_api/jwt_auth/middleware.py
@@ -73,6 +73,7 @@ class JWTMiddleware:
         auth_table: t.Type[BaseUser] = BaseUser,
         blacklist: JWTBlacklist = JWTBlacklist(),
         allow_unauthenticated: bool = False,
+        visible_paths: t.List[str] = [],
     ) -> None:
         """
         :param asgi:
@@ -88,6 +89,11 @@ class JWTMiddleware:
         :param allow_unauthenticated:
             By default the middleware rejects any requests with an invalid
             token.
+        :param visible_paths:
+            Useful for Swagger docs
+            (eg visible_paths=["/docs", "/openapi.json"])
+            Swagger docs can be viewed in a browser, but are
+            still token protected.
 
         """
         self.asgi = asgi
@@ -95,6 +101,7 @@ class JWTMiddleware:
         self.auth_table = auth_table
         self.blacklist = blacklist
         self.allow_unauthenticated = allow_unauthenticated
+        self.visible_paths = visible_paths
 
     def get_token(self, headers: dict) -> t.Optional[str]:
         """
@@ -129,12 +136,13 @@ class JWTMiddleware:
         is recognised, otherwise raise a 403 HTTP error.
         """
         allow_unauthenticated = self.allow_unauthenticated
+        visible_paths = self.visible_paths
 
         headers = dict(scope["headers"])
         token = self.get_token(headers)
         if not token:
             error = JWTError.token_not_found.value
-            if allow_unauthenticated:
+            if allow_unauthenticated or scope["path"] in visible_paths:
                 await self.asgi(
                     extend_scope(scope, {"user_id": None, "jwt_error": error}),
                     receive,
@@ -146,7 +154,7 @@ class JWTMiddleware:
 
         if await self.blacklist.in_blacklist(token):
             error = JWTError.token_revoked.value
-            if allow_unauthenticated:
+            if allow_unauthenticated or scope["path"] in visible_paths:
                 await self.asgi(
                     extend_scope(scope, {"user_id": None, "jwt_error": error}),
                     receive,
@@ -160,7 +168,7 @@ class JWTMiddleware:
             token_dict = jwt.decode(token, self.secret, algorithms=["HS256"])
         except jwt.exceptions.ExpiredSignatureError:
             error = JWTError.token_expired.value
-            if allow_unauthenticated:
+            if allow_unauthenticated or scope["path"] in visible_paths:
                 await self.asgi(
                     extend_scope(scope, {"user_id": None, "jwt_error": error}),
                     receive,
@@ -171,7 +179,7 @@ class JWTMiddleware:
                 raise HTTPException(status_code=403, detail=error)
         except jwt.exceptions.InvalidSignatureError:
             error = JWTError.token_invalid.value
-            if allow_unauthenticated:
+            if allow_unauthenticated or scope["path"] in visible_paths:
                 await self.asgi(
                     extend_scope(scope, {"user_id": None, "jwt_error": error}),
                     receive,
@@ -184,7 +192,7 @@ class JWTMiddleware:
         user = await self.get_user(token_dict)
         if user is None:
             error = JWTError.user_not_found.value
-            if allow_unauthenticated:
+            if allow_unauthenticated or scope["path"] in visible_paths:
                 await self.asgi(
                     extend_scope(scope, {"user_id": None, "jwt_error": error}),
                     receive,
--- a/tests/jwt_auth/test_jwt_middleware.py
+++ b/tests/jwt_auth/test_jwt_middleware.py
@@ -2,6 +2,7 @@ import datetime
 from unittest import TestCase
 
 import jwt
+from fastapi import FastAPI
 from piccolo.apps.user.tables import BaseUser
 from starlette.endpoints import HTTPEndpoint
 from starlette.exceptions import HTTPException
@@ -24,10 +25,15 @@ class EchoEndpoint(HTTPEndpoint):
         )
 
 
+FASTAPI_APP = FastAPI(title="Test visible paths")
+
 ECHO_APP = Router([Route("/", EchoEndpoint)])
 APP = JWTMiddleware(asgi=ECHO_APP, secret="SECRET")
 APP_UNAUTH = JWTMiddleware(
     asgi=ECHO_APP, secret="SECRET", allow_unauthenticated=True
+)
+APP_VISIBLE_PATHS = JWTMiddleware(
+    asgi=FASTAPI_APP, secret="SECRET", visible_paths=["/docs"]
 )
 
 
@@ -198,4 +204,14 @@ class TestJWTMiddleware(TestCase):
         self.assertDictEqual(
             response.json(),
             {"user_id": None, "jwt_error": JWTError.user_not_found.value},
+        )
+
+    def test_visible_paths(self):
+        client = TestClient(FASTAPI_APP)
+
+        response = client.get("/docs")
+        self.assertEqual(response.status_code, 200)
+        self.assertIn(
+            b"<title>Test visible paths - Swagger UI</title>",
+            response.content,
         )
--- a/docs/source/jwt/middleware.rst
+++ b/docs/source/jwt/middleware.rst
@@ -62,6 +62,67 @@ actions when an error occurs before rejecting the request.
 
 -------------------------------------------------------------------------------
 
+visible_paths
+~~~~~~~~~~~~~
+
+By default, if the JWT token is invalid then the HTTP request is rejected.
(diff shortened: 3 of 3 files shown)


piccolo_api/jwt_auth/middleware.py
```diff
@@ -129,12 +136,13 @@ async def __call__(self, scope, receive, send):
...
         """
         allow_unauthenticated = self.allow_unauthenticated
+        visible_paths = self.visible_paths
 
         headers = dict(scope["headers"])
         token = self.get_token(headers)
         if not token:
             error = JWTError.token_not_found.value
-            if allow_unauthenticated:
+            if allow_unauthenticated or visible_paths:
```
dantownsend, line 145: You need to be careful here because this is checking to the presence of `visible_paths`, and not whether it matches the current path.

sinisaos: Do you think `scope["path"] in visible_paths` is a better option?

sinisaos: I think checking presence of `visible_paths` (if is empty list is or not specified we got `Token not found` error) enough because we can make a GET request only for Swagger docs from browser on routes `/docs` and `/openapi.json` where we have to authorize the user if we want to communicate with the endpoints. All other routes must have a token in the header.

dantownsend: > Do you think scope["path"] in visible_paths is a better option?

Yes, something like that.

sinisaos: OK, I'll change that even though I think it might not be necessary (see previous comment) because if for example we pass a `/bands` route to `visible_paths`, we can't make a request to that route directly from the browser (we'll get `Token not found`). Basically we can only pass the Swagger docs routes `/docs` and `/openapi.json`. Sorry if I'm wrong.

dantownsend: If `visible_paths` matches the current path, then we should just let it bypass all other checks. I think this is the problem - it's still checking for a token etc, but we don't need to.

dantownsend: Put the check for `visible_paths` at the top level.

dantownsend:
```python
    async def __call__(self, scope, receive, send):
        """
        Add the user_id to the scope if a JWT token is available, and the user
        is recognised, otherwise raise a 403 HTTP error.
        """
        allow_unauthenticated = self.allow_unauthenticated
        visible_paths = self.visible_paths
        
        if scope['path'] in self.visible_paths:
            await self.asgi(scope, receive, send)
        
        # then all of the existing code
```

dantownsend: I'd maybe change `visible_paths` to something like `excluded_paths`. Meaning it's excluded from the checks.

sinisaos: Yeah. All the time I tested with
```python
auth_header = APIKeyHeader(name="Authorization")
private_app = FastAPI(dependencies=[Depends(auth_header)])
```
and with `auth_header` as dependency to the whole `private_app` only Swagger endpoints work in `visible_paths` because the authorization header dependency expected header on every request (that's why the `/bands` endpoint requested a token). I'm sorry. 
I will make the changes you suggested. If that's not enough, feel free to change anything.

dantownsend: Ok, interesting. I didn't consider the interaction with `APIKeyHeader`.

It might be best to attach `APIKeyHeader` to an `APIRouter` rather than the app.

sinisaos: Yes. We can do this at `APIRouter` level or if we use `FastAPIWrapper` we can do this
```python
auth_header = APIKeyHeader(name="Authorization")
private_app = FastAPI()

FastAPIWrapper(
    "/bands/",
    fastapi_app=private_app,
    piccolo_crud=PiccoloCRUD(Band, read_only=False),
    fastapi_kwargs=FastAPIKwargs(
        all_routes={"tags": ["Band"]},
        post={"dependencies": [Depends(auth_header)]},  # protected route
        put={"dependencies": [Depends(auth_header)]},  # protected route
        patch={"dependencies": [Depends(auth_header)]},  # protected route
        delete_single={
            "dependencies": [Depends(auth_header)]
        },  # protected route
    ),
)
```
But to be honest, I like it best at the application level, as we do in Piccolo Admin, because a protected (private) application should be private, and I think the best option is that only Swagger docs can be seen (for frontend developers as requested in issue) but in order to communicate with the endpoints, the user must be authorized which is provided with `APIKeyHeader`.

piccolo_api/jwt_auth/middleware.py
```diff
@@ -130,6 +137,10 @@ async def __call__(self, scope, receive, send):
         """
         allow_unauthenticated = self.allow_unauthenticated
 
+        if scope["path"] in self.excluded_paths:
```
dantownsend, line 140: We can enhance this quite easily to allow `*`. For example `foo/*`.

```python
for excluded_path in self.excluded_paths:
    if excluded_path.endswith('*'):
        if scope['path'].startswith(excluded_path.rstrip('*')):
            await self.asgi(scope, receive, send)
            return
    else:
        if scope['path'] == excluded_path:
            await self.asgi(scope, receive, send)
            return
```

It would be a really useful feature.
        
```

File: tests/jwt_auth/test_jwt_middleware.py [hunk 3]
```diff
@@ -199,3 +205,13 @@ def test_token_without_user_id(self):
             response.json(),
             {"user_id": None, "jwt_error": JWTError.user_not_found.value},
         )
+
+    def test_visible_paths(self):
```
dantownsend, line 209: We should update this to `test_excluded_paths`.

sinisaos: I forgot to change that. Thanks.

piccolo_api/jwt_auth/middleware.py
sinisaos: Great idea. With this we can exclude the entire paths of the app, for example for `private_app` we can exclude all paths like `private/*`. I only need to change this line 
```python
if scope['path'].startswith(excluded_path.rstrip('*')):
``` 
to 
```python
if excluded_path.startswith(excluded_path.rstrip("*")):
``` 
to have an effect.

piccolo_api/jwt_auth/middleware.py
```diff
@@ -137,9 +137,15 @@ async def __call__(self, scope, receive, send):
         """
         allow_unauthenticated = self.allow_unauthenticated
 
-        if scope["path"] in self.excluded_paths:
-            await self.asgi(scope, receive, send)
-            return
+        for excluded_path in self.excluded_paths:
+            if excluded_path.endswith("*"):
+                if excluded_path.startswith(excluded_path.rstrip("*")):
```
dantownsend, line 142: This isn't quite right - because it's comparing `excluded_paths` with `excluded_paths`.

sinisaos: But 
```python
if scope['path'].startswith(excluded_path.rstrip('*')):
```
does nothing.  If we have `/foo/*` we need check to `root_path` like this 
```python
if scope["root_path"].startswith(excluded_path.rstrip("/*")): 
```
to take some effect. Sorry if I don't understand well.

sinisaos:
```python
if scope['path'].startswith(excluded_path.rstrip('*')):
```
work only if we specified `excluded_paths=["*"]` not `excluded_paths=["/foo/*"]`. Did you think so?

dantownsend: I think the confusion is to do with out FastAPI splits the path between `path` and `root_path`.

For example, if I have the an `APIRouter` mounted at `/private`, and an endpoint mounted to that router at `/blog`, then `root_path` is `/private` and `path` is `blog`.

I think we need to combine them for it to work properly. So rather than just checking against `path` it's something like `urllib.parse.urljoin(request.scope['root_path'], request.scope['path'])`.

dantownsend: There's also `raw_path`:

https://asgi.readthedocs.io/en/latest/specs/www.html#http-connection-scope

dantownsend: This discussion gives some more context https://github.com/django/asgiref/issues/229

sinisaos: `raw_path` did the trick. If we use `raw_path` like this
```python
if scope["raw_path"].decode("utf-8").startswith(excluded_path.rstrip("*")):
```
both `excluded_paths=["*"]` and `excluded_paths=["/foo/*"]` works.

tests/jwt_auth/test_jwt_middleware.py
```diff
@@ -39,7 +39,7 @@ def wildcard_route():
     asgi=ECHO_APP, secret="SECRET", allow_unauthenticated=True
 )
 APP_EXCLUDED_PATHS = JWTMiddleware(
-    asgi=fastapi_app, secret="SECRET", excluded_paths=["/docs", "/*"]
+    asgi=fastapi_app, secret="SECRET", excluded_paths=["/docs", "*"]
```
dantownsend, line 42: We should remove `"*"` as this lets everything through.

sinisaos: That will decrease the coverage. If we remove that, we need to add another FastAPI application to test the route with wildcards like this
```python
fastapi_app_wildcard = FastAPI()

@fastapi_app_wildcard.get("/")
def wildcard_route():
    return "Wildcard route test"

APP_EXCLUDED_PATHS_WILDCARD = JWTMiddleware(
    asgi=fastapi_app_wildcard,
    secret="SECRET",
    excluded_paths=["*"],
)

def test_excluded_paths_wildcards(self):
    client = TestClient(APP_EXCLUDED_PATHS_WILDCARD)

    response = client.get("/")
    self.assertEqual(response.status_code, 200)
    self.assertIn(
        b"Wildcard route test",
        response.content,
    )
```
Is that OK?

dantownsend: I was thinking something like this:

```python
fastapi_app_wildcard = FastAPI()

@fastapi_app_wildcard.get("/")
def home_root():
    return "Root"

@fastapi_app_wildcard.get("/path/a/")
def sub_root():
    return "Sub route"

APP_EXCLUDED_PATHS_WILDCARD = JWTMiddleware(
    asgi=fastapi_app_wildcard,
    secret="SECRET",
    excluded_paths=["/path/*"],
)

def test_excluded_paths_wildcards(self):
    client = TestClient(APP_EXCLUDED_PATHS_WILDCARD)

    response = client.get("/")
    # Requires a token
    self.assertEqual(response.status_code, 403)

    # Is an excluded path, so doesn't need a token
    response = client.get("/path/a/")
    self.assertEqual(response.status_code, 200)
```

sinisaos: OK. I will add your tests with one change
```python
def test_excluded_paths_wildcards(self):
    client = TestClient(APP_EXCLUDED_PATHS_WILDCARD)

    with self.assertRaises(HTTPException):
        response = client.get("/")
        # Requires a token
        self.assertEqual(response.status_code, 403)

    # Is an excluded path, so doesn't need a token
    response = client.get("/path/a/")
    self.assertEqual(response.status_code, 200)

```

sinisaos:
@dantownsend What is the problem that this is not merged, since it is practically the same as TokenAuth `excluded_paths` which is merged? Do I need to change something?

dantownsend:
@sinisaos I think it just slipped through the cracks. I can't remember if we tweaked the docs or something on the other PR. Will have a look later on.

sinisaos:
@dantownsend Great. Thanks

PR #227, opened 2023-04-19
add excluded_paths to token auth

sinisaos:
Added `excluded_paths` to token authentication. It is useful when we want to have Swagger docs in a token protected application.


docs/source/token_auth/middleware.rst
```diff
@@ -65,6 +65,68 @@ You'll have to run the migrations for this to work correctly.
...
+    auth_header = APIKeyHeader(name="Authorization")
+    private_app = FastAPI(dependencies=[Depends(auth_header)])
+
+    protected_app = AuthenticationMiddleware(
+        private_app,
+        backend=TokenAuthBackend(
+            PiccoloTokenAuthProvider(),
+            excluded_paths=["/docs", "/openapi.json"],
+        ),
+    )
```
dantownsend, line 114: We can do this, which might makes things slightly easier to understand:

```python
from fastapi.middleware import Middleware

private_app = FastAPI(
    dependencies=[Depends(auth_header)],
    middleware=[
        Middleware(
            AuthenticationMiddleware,
            backend=TokenAuthBackend(
                PiccoloTokenAuthProvider(),
                excluded_paths=["/docs", "/openapi.json"],
            )
        )
    ]
)
```

What do you think?

piccolo_api/token_auth/middleware.py
```diff
@@ -84,9 +84,11 @@ class TokenAuthBackend(AuthenticationBackend):
     def __init__(
         self,
         token_auth_provider: TokenAuthProvider = DEFAULT_PROVIDER,
+        excluded_paths: t.List[str] = [],
     ):
         super().__init__()
         self.token_auth_provider = token_auth_provider
+        self.excluded_paths = excluded_paths
```
dantownsend, line 91: This is a very minor nitpick, but this is marginally better:

```python
    def __init__(
        self,
        token_auth_provider: TokenAuthProvider = DEFAULT_PROVIDER,
        excluded_paths: t.Optional[t.List[str]] = None,
    ):
        super().__init__()
        self.token_auth_provider = token_auth_provider
        self.excluded_paths = excluded_paths or []
```

I try and avoid having lists as default values because it can sometimes lead to bugs (if we mutate that list it effects all instances).

dantownsend:
@sinisaos This is great, thanks!

piccolo_api/token_auth/middleware.py
dantownsend: Or maybe t.Sequence instead of t.List, as we don't explicitly need it to be a list.

docs/source/token_auth/middleware.rst
sinisaos: This is just an example of usage. I wanted to be consistent with `JWTMiddleware` in terms of its implementation and usage as this is similar. If you want I can change it in docs.

piccolo_api/token_auth/middleware.py
sinisaos: OK. I will change that.

dantownsend:
@sinisaos I added some images, and used Sphinx's `literalinclude` so we can run the example app to make sure it works.

We need to make sure the docs and features are consistent between the different auth backends, so will have to do that once the other PR is merged in.

PR #230, opened 2023-05-06
Update Starlette URL in README

xEricL:
https://starlette.io does not resolve to a valid website. The correct url includes `www.` (https://www.starlette.io)

image


dantownsend:
@xEricL Good catch - thanks!

PR #238, opened 2023-07-07
pin FastAPI and Pydantic 1 version

sinisaos:
Related to [this issue](https://github.com/piccolo-orm/piccolo_admin/issues/313)


dantownsend:
I'm going to see if `pydantic[email]>=1.6,<2.0` works, just in case they release a critical bug fix version to v1, we had better not pin to a single version.

PR #245, opened 2023-07-27
pydantic v2 support

sinisaos:
Related to [https://github.com/piccolo-orm/piccolo_admin/issues/313](https://github.com/piccolo-orm/piccolo_admin/issues/313).

The tests will fail until we pin it to the first version of Piccolo with Pydantic V2 support, when it is released.

### DIFF
--- a/piccolo_api/crud/endpoints.py
+++ b/piccolo_api/crud/endpoints.py
@@ -24,7 +24,6 @@ from piccolo.query.methods.delete import Delete
 from piccolo.query.methods.select import Select
 from piccolo.table import Table
 from piccolo.utils.encoding import dump_json
-from pydantic.error_wrappers import ValidationError
 from starlette.requests import Request
 from starlette.responses import JSONResponse, Response
 from starlette.routing import Route, Router
@@ -38,7 +37,7 @@ from piccolo_api.crud.hooks import (
 )
 
 from .exceptions import MalformedQuery, db_exception_handler
-from .serializers import Config, create_pydantic_model
+from .serializers import create_pydantic_model
 from .validators import Validators, apply_validators
 
 if t.TYPE_CHECKING:  # pragma: no cover
@@ -315,7 +314,7 @@ class PiccoloCRUD(Router):
             include_readable=include_readable,
             include_columns=include_columns,
             model_name=f"{self.table.__name__}Output",
-            nested=nested,
+            nested=nested,  # type:ignore
         )
 
     @property
@@ -354,11 +353,13 @@ class PiccoloCRUD(Router):
             include_readable=include_readable,
             include_columns=include_columns,
             model_name=f"{self.table.__name__}Item",
-            nested=nested,
+            nested=nested,  # type:ignore
         )
         return pydantic.create_model(
             str(self.table.__name__) + "Plural",
-            __config__=Config,
+            __config__=pydantic.config.ConfigDict(
+                arbitrary_types_allowed=True
+            ),
             rows=(t.List[base_model], None),
         )
 
@@ -367,7 +368,7 @@ class PiccoloCRUD(Router):
         """
         Return a representation of the model, so a UI can generate a form.
         """
-        return JSONResponse(self.pydantic_model.schema())
+        return JSONResponse(self.pydantic_model.model_json_schema())
 
     ###########################################################################
 
@@ -713,7 +714,7 @@ class PiccoloCRUD(Router):
         """
         fields = params.fields
         if fields:
-            model_dict = self.pydantic_model_optional(**fields).dict()
+            model_dict = self.pydantic_model_optional(**fields).model_dump()
             for field_name in fields.keys():
                 value = model_dict.get(field_name, ...)
                 if value is ...:
@@ -865,7 +866,7 @@ class PiccoloCRUD(Router):
             include_readable=include_readable,
             include_columns=tuple(visible_fields),
             nested=nested,
-        )(rows=rows).json()
+        )(rows=rows).model_dump_json()
         return CustomJSONResponse(json, headers=headers)
 
     ###########################################################################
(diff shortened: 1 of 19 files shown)


piccolo_api/crud/endpoints.py
```diff
@@ -315,7 +314,7 @@ def _pydantic_model_output(
             include_readable=include_readable,
             include_columns=include_columns,
             model_name=f"{self.table.__name__}Output",
-            nested=nested,
+            nested=nested,  # type:ignore
```
dantownsend, line 317: What's the type error here?

piccolo_api/fastapi/endpoints.py
```diff
@@ -413,8 +413,13 @@ def modify_signature(
...
 
-        for field_name, _field in model.__fields__.items():
-            type_ = _field.outer_type_
+        for field_name, _field in model.model_fields.items():
+            # get type of field since field.outer_type_ is
+            # deprecated in Pydantic V2 (we use that for arrays
+            # in the OpenAPI docs). Using vars(), but we can
+            # also use _field.annotation.__dict__ (if that's better)
+            field_annotation = vars(_field.annotation)
+            type_ = field_annotation["__args__"][0]
```
dantownsend, line 422: I'm not sure what the best approach is here. I know we used `outer_type` because if it's `list[int]` it returns `list`.

What does `_field.annotation` give us?

tests/crud/test_crud_endpoints.py
```diff
@@ -493,33 +511,31 @@ class Rating(Enum):
         self.assertEqual(
             response.json(),
             {
-                "title": "ReviewIn",
-                "type": "object",
+                "help_text": None,
+                "primary_key_name": "id",
```
dantownsend, line 515: So most of these changes are because Pydantic v2 changed the order of fields in the response?

sinisaos: Yes. In Pydantic v2 the schema fields are sorted in alphabetical order.

piccolo_api/fastapi/endpoints.py
sinisaos: For array field of strings `_field.annotation` is `typing.Optional[typing.List[str]]` but result is wrong

without_args


For array field of strings `field_annotation["__args__"][0]` is `typing.List[str]` which is correct

with_args

I don't know how to solve it otherwise, if you have any other idea feel free to write here.

sinisaos:
@dantownsend Any updates or new thoughts on this? I think the Piccolo API and Piccolo Admin are equally important parts, as the ORM, in the Piccolo ecosystem and it would be nice if they also have support for Pydantic V2 as soon as possible. Libraries without this support fall out of use, which would be a real shame, because Piccolo is one of the best-maintained, feature-rich libraries out there.

dantownsend:
@sinisaos Yeah, I'm running a bit behind of these PRs - has been a busy couple of weeks. But I plan on getting them out asap, because as you say we need them all released, and not just Piccolo.

piccolo_api/crud/endpoints.py
dantownsend: OK, I see the problem. I think it has always been an issue, but before we wrapped `create_pydantic_model` in `@lru_cache`, which probably meant mypy couldn't see it. I have a fix.

tests/crud/test_crud_endpoints.py
```diff
@@ -282,6 +279,24 @@ def test_patch_fails(self):
...
+        client = TestClient(PiccoloCRUD(table=Movie, read_only=False))
+
+        movie = Movie(name="Star Wars", rating=93)
+        movie.save().run_sync()
+
+        response = client.patch(
+            f"/{movie.id}/",
+            json={"name": 95, "rating": "95"},
+        )
+        self.assertIn("validation error", str(response.content))
```
dantownsend, line 295: Thanks for adding this.

In Pydantic v2, the validation error we get back is pretty verbose / technical:

```
b'1 validation error for MovieOptional\nname\n  Input should be a valid string [type=string_type, input_value=95, input_type=int]\n    For further information visit https://errors.pydantic.dev/2.3/v/string_type'
```

I guess it's OK.

tests/crud/test_crud_endpoints.py
```diff
@@ -282,6 +279,24 @@ def test_patch_fails(self):
...
+        self.assertEqual(response.status_code, 400)
+
+        self.assertEqual(Movie.count().run_sync(), 1)
```
dantownsend, line 298: I'm going to change this assertion - because we want to make sure the existing row didn't change.

dantownsend (approved):
Looks great 👍 

Sorry it took me so long - I needed a day off to devote to it, so I could wrap my head around some of the Pydantic changes.

piccolo_api/fastapi/endpoints.py
sinisaos: I just tried this and the `_get_type` function doesn't work as it should. The problem remains the same because the `Array` column does not display correct in the `OpenAPI` document filters. Now it's like this.

without_args

If I change the `_get_type` function to something like this
```python
def _get_type(type_: t.Type) -> t.Type:
    """
    Extract the inner type from an optional if necessary, otherwise return
    the type as is.
    """
    args = t.get_args(type_)
    return args[0]
```
the result is correct and looks like this.

with_args

Can you check that in case I'm wrong or doing something wrong.

dantownsend: You're right - I can see the problem when trying to use the `Director.years_nominated` filter in the Swagger docs.

Will investigate.

dantownsend: It's a bit odd, but `t.get_args` returns this:  `(typing.List[int], <class 'NoneType'>)` and I was expecting `(typing.List[int], None)`.

I need to work out what `NoneType` is vs `None`.

dantownsend: So `type(None)` is `NoneType`. I've changed it to check for `None` and `NoneType`, and it should work.

Will add some tests.

PR #248, opened 2023-07-29
fix #246

### LINKED ISSUE #246: `schema_extra` param not passed to `pydantic_model_{output | optional | plural}` methods
https://github.com/piccolo-orm/piccolo_api/blob/e2f8ce83a29234d0e6a79b4add044be71b904637/piccolo_api/crud/endpoints.py#L294-L304
Why is schema_extra not passed to the following methods as in this method above?
```
pydantic_model_output
pydantic_model_optional
pydantic_model_plural
```
When I use my pydantic basic config it only works for `ModelIn` and others don't
```python
def to_camel(string: str) -> str:
    string_split = string.split("_")
    return string_split[0] + "".join(word.capitalize() for word in string_split[1:])

class MyConfig(pydantic.BaseConfig):
    alias_generator = to_camel
    allow_population_by_field_name = True

chat_crud = PiccoloCRUD(
    table=Message,
    read_only=False,
    max_joins=2,
    schema_extra={
        'pydantic_config_class': MyConfig
    }
)

```
It is convenient for my frontend that every input and output field is in `camelCase`. As another solution, I could name each field in my `Table` as camelCase, but then I would have to change my backend structure.

hoosnick:
passed the `schema_extra` parameter to each method where `create_pydantic_model` is called in PiccoloCRUD

### DIFF
--- a/piccolo_api/crud/endpoints.py
+++ b/piccolo_api/crud/endpoints.py
@@ -315,7 +315,7 @@ class PiccoloCRUD(Router):
             include_readable=include_readable,
             include_columns=include_columns,
             model_name=f"{self.table.__name__}Output",
-            nested=nested,
+            nested=nested, **self.schema_extra,
         )
 
     @property
@@ -337,6 +337,7 @@ class PiccoloCRUD(Router):
             include_default_columns=True,
             all_optional=True,
             model_name=f"{self.table.__name__}Optional",
+            **self.schema_extra,
         )
 
     def pydantic_model_plural(
@@ -355,6 +356,7 @@ class PiccoloCRUD(Router):
             include_columns=include_columns,
             model_name=f"{self.table.__name__}Item",
             nested=nested,
+            **self.schema_extra,
         )
         return pydantic.create_model(
             str(self.table.__name__) + "Plural",


piccolo_api/crud/endpoints.py
```diff
@@ -337,6 +337,7 @@ def pydantic_model_optional(self) -> t.Type[pydantic.BaseModel]:
             include_default_columns=True,
             all_optional=True,
             model_name=f"{self.table.__name__}Optional",
+            **self.schema_extra,
```
sinisaos, line 340: Sorry for the slow reply and sorry if I missed your point and doing something wrong, but this will break the `PATCH` request. I tried something like this
```python
class Message(Table, db=DB):
    message_title = Varchar(length=100)
    message_content = Varchar()

app = FastAPI(debug=True)

def to_camel(string: str) -> str:
    string_split = string.split("_")
    return string_split[0] + "".join(
        word.capitalize() for word in string_split[1:]
    )

class MyConfig(pydantic.BaseConfig):
    alias_generator = to_camel
    allow_population_by_field_name = True

FastAPIWrapper(
    root_url="/message/",
    fastapi_app=app,
    piccolo_crud=PiccoloCRUD(
        table=Message,
        read_only=False,
        schema_extra={"pydantic_config_class": MyConfig},
    ),
)
```
and I get a `400` error on the `PATCH` request `Unrecognized keys - {'messageContent', 'messageTitle'}` (using `curl`) because the `camelCase` field is not recognized. 
Also `PUT` request get `500` error (with that Pydantic configuration of `camelCase` in `schema_extra`) with mesage `AttributeError: type object 'Message' has no attribute 'messageTitle' `
Can you confirm that?

hoosnick: Don't worry, better late than never xD. Yes, I saw. I didn't quite understand these 3 data variables/params in the following lines:

https://github.com/piccolo-orm/piccolo_api/blob/e2f8ce83a29234d0e6a79b4add044be71b904637/piccolo_api/crud/endpoints.py#L1104-L1132

```
cleaned_data
data
model
```
If I change the code like below we can achieve something:
```python
values = {
  getattr(cls, key): getattr(model, key) for key in model.dict().keys()
}
```
ie new problems:
image
**Can you test?**

sinisaos: I will try to explain (as I understand it) and I hope it will be clear enough :).

> cleaned_data - payload handling null value (convert null as string in payload to `None`)
> data - payload data
> model - Pydantic model

The tests fail on the `PATCH` request (partial update) because `data` dict are not the same as `model.dict()`, because the model uses [pydantic_model_optional](https://github.com/piccolo-orm/piccolo_api/blob/e2f8ce83a29234d0e6a79b4add044be71b904637/piccolo_api/crud/endpoints.py#L330) which puts the optional fields as `None` and if we patch only one field we need `data from payload` not from `model`.
Example of payload data:
```
{'name': 'Star Wars: A New Hope'} <-- data dict
{'id': None, 'name': 'Star Wars: A New Hope', 'rating': None} <-- model.dict()
```
and because of that we get error `422` in `patch` tests if we use
```python
values = {
  getattr(cls, key): getattr(model, key) for key in model.dict().keys()
}
```

Issue #249, opened 2023-07-31
Piccolo Admin API docs are not rendering


sinisaos:
Piccolo Admin API docs are not rendering because we need to upgrade `Swagger UI` from version 4 to version 5.

![error](https://github.com/piccolo-orm/piccolo_api/assets/30960668/4bb7fd2d-83bd-4595-a6df-a349abfbf4c1)

PR #250, opened 2023-07-31
upgrade Swagger UI

sinisaos:
Related to #249


dantownsend:
@sinisaos Thanks. I had a quite look at the release notes for Swagger UI:

https://github.com/swagger-api/swagger-ui/releases/tag/v5.0.0

I think it's backwards compatible - so by upgrading, it will hopefully still work with old versions of the OpenAPI spec.

PR #252, opened 2023-08-01
Make Swagger UI version configurable

dantownsend:
Related to https://github.com/piccolo-orm/piccolo_api/issues/249

@sinisaos I guess we may as well make the Swagger UI version configurable.


sinisaos (approved):
This is great because with these changes we will be able to change the version in one place if needed.

PR #254, opened 2023-09-06
fix `_get_type` - check for `NoneType`

dantownsend:
Related to https://github.com/piccolo-orm/piccolo_api/pull/245#discussion_r1316021624


sinisaos (approved):
@dantownsend I can confirm that these changes are working as they should.

dantownsend:
@sinisaos Thanks for checking 👍 

Just added one more test, but seems to failing.

dantownsend:
@sinisaos Ah, great catch! I'll change it.

Issue #267, opened 2024-02-22
RateLimitingMiddleware tests failed


sinisaos:
Scope client is set to `None` in [Starlette>=0.35.0](https://www.starlette.io/release-notes/#0350) testclient. Due to this change, the test for `RateLimitingMiddleware` failed in Piccolo API and Piccolo Admin. We can fix it by changing rate limiter middleware [dispatch method](https://github.com/piccolo-orm/piccolo_api/blob/8b5cc865dc4172a466bcba5e1208a32fdc6bf027/piccolo_api/rate_limiting/middleware.py#L159) like this
```python
async def dispatch(
        self, request: Request, call_next: RequestResponseEndpoint
) -> Response:
    request_client = request.url.netloc
    if not request_client:
        # If we can't get the client, we have to reject the request.
        return Response(
            content="Client host can't be found.", status_code=400
        )

    identifier = request_client
    try:
        self.rate_limit.increment(identifier)
    except RateLimitError:
        return Response(content="Too many requests", status_code=429)
    return await call_next(request)
```
 or based on [this discussion](https://github.com/encode/starlette/pull/2377) you can change the testing method. I leave this decision to you.

dantownsend:
@sinisaos Interesting - so it just effects the tests? If it broke `RateLimitingMiddleware` that would be bad, as it's used by Piccolo Admin.

dantownsend:
I think I'd rather fix the test somehow, as the problem is related to the changes to `TestClient`.

I sometimes use the `httpx.Client` directly instead of TestClient, so maybe this doesn't have the same problem.

dantownsend:
@sinisaos Thanks for reporting this. I have a fix for now.

sinisaos:
@dantownsend I don't think all these changes to the Piccolo API and Admin were necessary because if the Starlette testclient reverts back to provide scope client (there is a PR for that), all the tests will pass again. Apologies if I wasn't clear enough in my previous comment, **but the rate limit is only broken in tests, not in real use**. With the changes I suggested in the comment, there was no need to change the code (except the [dispatch](https://github.com/piccolo-orm/piccolo_api/issues/267#issue-2148558360) method) in the Piccolo API and Piccolo Admin, regardless of whether Starlette revert client scope to testclient or leaves it to `None` as it is now. But it doesn't matter because you already make all these changes in both repositories and they work.

dantownsend:
[...]

PR #268, opened 2024-02-22
Fix rate limiting test

dantownsend:
Resolves https://github.com/piccolo-orm/piccolo_api/issues/267


dantownsend:
Looks like there are fresh linter errors ...

PR #272, opened 2024-03-15
docs: Fix typo

jrycw:
Based on the comment, it seems the function name should be `set_movie_rating_20`.


dantownsend:
@jrycw Good catch - thanks! 👍

PR #280, opened 2024-03-30
Fix the CSP regression

Skelmis:
The previous version made a change to the CSP which results in an incorrect CSP policy being set, which leads to it being ignored by browsers.


dantownsend:
Oops, good catch! Some tests are failing because we need to add the colon to the expected result.

PR #284, opened 2024-05-16
Use secure and httponly csrf cookie

Astra-X:


piccolo_api/csrf/middleware.py
```diff
@@ -120,6 +120,8 @@ async def dispatch(
                     self.cookie_name,
                     token,
                     max_age=self.max_age,
+                    httponly=True,
+                    secure=True,
```
dantownsend, line 124: The only reason this was currently set to `False` is because when using Piccolo Admin locally, most people wouldn't run it under HTTPS so it would stop working. I think a good compromise would be to have this configurable. For example, we have a `production` argument with Session auth:

https://github.com/piccolo-orm/piccolo_api/blob/f3e17a610a61c2cbedf57cf154a339f85ba8d84e/piccolo_api/session_auth/endpoints.py#L330

dantownsend (commented):
Thanks for the PR.

I've left some comments about why the values are currently set as they - happy to discuss further.

piccolo_api/csrf/middleware.py
AmazingAkai: Hmm, this needs to be updated in piccolo admin as well, right?

dantownsend: Yeah, that's right. We would pass `production` in here: https://github.com/piccolo-orm/piccolo_admin/blob/5e8e96d56656cbf7283b63f8729d0748eb895867/piccolo_admin/endpoints.py#L439

AmazingAkai:
@dantownsend Is this alright now?

PR #302, opened 2024-10-04
replace 403 with 401 status code

### LINKED ISSUE #301: `401` vs `403` in JWTAuth
I think we should change the `status_code` from `403` to `401` if [credentials are missing or incorrect](https://supertokens.com/blog/http-error-codes-401-vs-403). In [TokenAuth](https://github.com/piccolo-orm/piccolo_api/blob/d6ed009d36d0f965aee3597cde956dd5dc4491e6/piccolo_api/token_auth/endpoints.py#L55-L69) and [SessionAuth](https://github.com/piccolo-orm/piccolo_api/blob/d6ed009d36d0f965aee3597cde956dd5dc4491e6/piccolo_api/session_auth/endpoints.py#L95) we use the proper `401` `status_code`. Sorry if  I'm wrong.

sinisaos:
Resolves #301

### DIFF
--- a/docs/source/jwt/middleware.rst
+++ b/docs/source/jwt/middleware.rst
@@ -11,7 +11,7 @@ Setup
 -----
 
 ``JWTMiddleware`` wraps an ASGI app, and ensures a valid token is passed in the header.
-Otherwise a 403 error is returned. If the token is valid, the corresponding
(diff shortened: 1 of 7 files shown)


dantownsend:
Looks great, thanks 👍

PR #303, opened 2024-11-09
Adds Azure Blob Storage media backend

terricain:
Adds a MediaStorage plugin for Azure Blob Storage


dantownsend:
@terricain Thanks for this!

piccolo_api/media/azure.py
```diff
@@ -0,0 +1,432 @@
...
+
+        if not self.sign_urls:
+            return (
+                f"https://{self.storage_account_name}.blob.core.windows.net/"
+                f"{self.container_name}/{blob_name}"
+            )
+
+        _, blob_service_client = self.get_client()
+
+        now = datetime.datetime.now(datetime.UTC)
```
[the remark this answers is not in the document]
terricain: Ah yep, makes sense 😄. Should be fixed now

piccolo_api/media/azure.py
```diff
@@ -0,0 +1,432 @@
...
+                "AZURE_STORAGE_CONNECTION_STRING"
+            ]
+        if self.connection_string:
+            self.connection_string_parts = dict(
+                pair.split("=", 1)
+                for pair in self.connection_string.split(";")
+            )
+            if "AccountKey" not in self.connection_string_parts:
+                sys.exit(
+                    "Specified connection string did not contain AccountKey"
```
dantownsend, line 165: Maybe we just raise a `ValueError` here

piccolo_api/media/azure.py
```diff
@@ -0,0 +1,432 @@
...
+        self, file_key: str, root_url: str, user: t.Optional[BaseUser] = None
+    ) -> str:
+        """
+        A sync wrapper around :meth:`generate_file_url`.
+        """
+        blob_name = self._prepend_folder_name(file_key)
+
+        if not self.sign_urls:
+            return (
+                f"https://{self.storage_account_name}.blob.core.windows.net/"
```
dantownsend, line 271: Is it worth having `blob.core.windows.net` as a class variable, or someone can override this somehow if needed? Just in case there are other services which are API compatible with Azure storage.

piccolo_api/media/azure.py
```diff
@@ -0,0 +1,432 @@
...
+
+
+class AzureMediaStorage(MediaStorage):
+    def __init__(
+        self,
+        column: t.Union[Text, Varchar, Array],
+        storage_account_name: str,
+        container_name: str,
+        folder_name: t.Optional[str] = None,
+        connection_kwargs: t.Optional[t.Dict[str, t.Any]] = None,
```
dantownsend, line 43: We don't really use `connection_kwargs` for much at the moment - it is just being used as a way of passing a connection string.

Should we pass the `connection_kwargs` to `DefaultAzureCredential`? I know the user can set most of the options via environment variables, but it might be convenient to pass them in directly.

https://learn.microsoft.com/en-us/python/api/azure-identity/azure.identity.defaultazurecredential?view=azure-python#constructor

In which case, connection string, could be a separate argument, or:

```python
connection_kwargs: t.Optional[t.Dict[str, t.Any]] | str = None
```

If it's a string, we treat it as a connection string.

What do you think?

piccolo_api/media/azure.py
```diff
@@ -0,0 +1,432 @@
...
+        self.connection_string = None
+        self.connection_string_parts: t.Dict[str, str] = {}
+        if "connection_string" in self.connection_kwargs:
+            self.connection_string = self.connection_kwargs[
+                "connection_string"
+            ]
+        elif "AZURE_STORAGE_CONNECTION_STRING" in os.environ:
+            self.connection_string = os.environ[
+                "AZURE_STORAGE_CONNECTION_STRING"
+            ]
```
dantownsend, line 157: Not really a big deal, but we can use the walrus operator to simplify this a little bit.

```python
if connection_string := (
    self.connection_kwargs.get("connection_string")
    or os.environ.get("AZURE_STORAGE_CONNECTION_STRING")
):
    self.connection_string = connection_string

```

piccolo_api/media/azure.py
```diff
@@ -0,0 +1,432 @@
...
+        Returns an Azure client.
+        """
+        if self.connection_string:
+            blob_service_client = BlobServiceClient.from_connection_string(
+                self.connection_string
+            )
+        else:
+            account_url = (
+                f"https://{self.storage_account_name}.blob.core.windows.net"
+            )
```
dantownsend, line 187: We reference this URL in quite a few places - what do you think about setting it in the constructor, so we just do `self.account_url = f'https://{self.storage_account_name}.blob.core.windows.net'` and then reference `self.account_url` in the rest of the code?

piccolo_api/media/azure.py
```diff
@@ -0,0 +1,432 @@
...
+
+            blob_service_client = BlobServiceClient(
+                account_url, credential=default_credential
+            )
+
+        container_client = blob_service_client.get_container_client(
+            self.container_name
+        )
+
+        return container_client, blob_service_client
```
dantownsend, line 200: I don't know a huge amount about Azure, and the differences between the two clients. We could split this into two separate methods ... one for each client type ... not sure. We can leave it as is for now.

dantownsend (commented):
Thanks a lot for this - sorry it took me a while to get around to reviewing it.

I think it looks great, just left some minor comments 👍

piccolo_api/media/azure.py
terricain: Yep

piccolo_api/media/azure.py
terricain: Yeah good shout, its rare but it _would_ change if dealing with GovClouds

piccolo_api/media/azure.py
terricain: I thought about it, I think instead of having a union of an optional dictionary or string, leave it as an optional dictionary and if `'connection_string'` is not present, pass the kwargs to DefaultCredential

piccolo_api/media/azure.py
terricain: Managed to simplify this somewhat taking into account the previous message.

piccolo_api/media/azure.py
terricain: We could just return the blob_service_client (which we have to instantiate differently based on how you're authing) and then call something like `self.get_client().get_container_client(self.container_name)` if needed but we need the container client 90% of the time, so I figured what we have now is slightly simpler. Up to you.

terricain:
Sorry for the wait, been a bit busy, have addressed the comments, let me know what you think.

terricain:
@dantownsend this is ready to be merged I think

PR #305, opened 2025-05-23
Update dependecies

sinisaos:
Update dependecies and fix a few mypy errors.

### DIFF
--- a/piccolo_api/fastapi/endpoints.py
+++ b/piccolo_api/fastapi/endpoints.py
@@ -121,7 +121,7 @@ class FastAPIWrapper:
 
         self.alias = f"{piccolo_crud.table._meta.tablename}__{id(self)}"
 
-        global ANNOTATIONS
+        # global ANNOTATIONS
         ANNOTATIONS[self.alias]["ModelIn"] = self.ModelIn
         ANNOTATIONS[self.alias]["ModelOut"] = self.ModelOut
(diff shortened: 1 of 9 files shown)


sinisaos:
@dantownsend I reverted the dependency changes. I only need to change one type annotation in `auth_token` `tables.py` because without that change Mypy complains and linters don't pass

requirements/doc-requirements.txt
```diff
@@ -1,3 +1,3 @@
 Sphinx==7.2.6
 piccolo-theme==0.21.0
-sphinx-autobuild==2024.2.4
+sphinx-autobuild==2024.2.4
```
dantownsend, line 3: What changed here? Just whitespace?

sinisaos:
@dantownsend I reverted the changes. I hope that's okay.

piccolo_api/fastapi/endpoints.py
```diff
@@ -121,7 +121,6 @@ def __init__(
 
         self.alias = f"{piccolo_crud.table._meta.tablename}__{id(self)}"
 
-        global ANNOTATIONS
```
dantownsend, line 124: We need to keep this - we need to make sure the global ANNOTATIONS are updated. Is it causing mypy errors? If so, lets ignore the error.

dantownsend: I put a `noqa` comment next to it.

dantownsend:
@sinisaos Thanks!

